Issue #13, opened 2021-08-06
Rename module names to start with `com.microsoft.`


sormuras:
As requestd by @brunoborges https://twitter.com/brunoborges/status/1423503746995298304 😉

Rename module names to start with `com.microsoft.`, in order to follow the reverse-DNS style pattern. Find more information and detailed reasonings here: https://blog.joda.org/2017/04/java-se-9-jpms-module-naming.html

Nice side-effect: their Maven coordinates will be automatically mapped by https://github.com/sormuras/modules

karianna:
@sormuras - Is it just the blog update to go now you think?

sormuras:
Yes.

Perhaps also the sample?

https://github.com/microsoft/gctoolkit/blob/main/sample/src/main/java/module-info.java#L10

But that's debatable, as that modular application (the module `gctoolkit.sample`) is not intended to be reused.

karianna:
> Yes.
> 
> Perhaps also the sample?
> 
> https://github.com/microsoft/gctoolkit/blob/main/sample/src/main/java/module-info.java#L10
> 
> But that's debatable, as that modular application (the module `gctoolkit.sample`) is not intended to be reused.

@dsgrieve - Thoughts on this?

sormuras:
Fine. Let's rename the sample module from `gctoolkit.sample` to `com.microsoft.gctoolkit.sample`. Want me to file a PR?

dsgrieve:
@sormuras I don't think you need a new PR for the sample. Just update this one.

sormuras:
This is an issue -- PR #15 was already merged and the underlying branch deleted.

dsgrieve:
Oh my. How did I not notice that!? If you would like to create a new PR, that would be awesome!

sormuras:
After PR #53 was (reviewed, approved, and) merged, only the blog post needs an update regarding new module names.

> - [ ] Update module snippet here https://devblogs.microsoft.com/java/introducing-microsoft-gctoolkit

karianna:
@brunoborges - would you be OK to update the blog or is that a @kirk-microsoft task?

brunoborges:
> @brunoborges - would you be OK to update the blog or is that a @kirk-microsoft task?

I can do that, yes.

PR #15, opened 2021-08-06
Prepend `com.microsoft.` to all main modules

sormuras:
Addresses first task of
- #13

I left the internal test modules untouched, as they are not meant to re-used by others, i.e. they are not published to Maven Central.


karianna (approved):
This change LGTM - If GCToolkit ever has its own domain this would change again, but it is currently an MSFT OSS project.

dsgrieve (approved):

PR #16, opened 2021-08-06
Make use of `javac`'s `--release` option

sormuras:
Prior to this commit the `-source` and `-target` option were used in two different ways.

This commit merges the two (old) options into the new one and also removes a configuration duplication.


karianna (approved):
LGTM

PR #27, opened 2021-08-10
Update Docs to improve understanding for first time contributors

karianna:


README.md
```diff
@@ -54,31 +55,26 @@ TODO Add some context as to why we need this and where it should go.
 <dependency>
     <groupId>com.microsoft.gctoolkit</groupId>
     <artifactId>api</artifactId>
-    <version>2.0.1-SNAPSHOT</version>
+    <version>2.0.1</version>
```
dsgrieve, line 58: I think the version here should refer to the artifact in github packages, not the version in source.

sample/src/main/java/com/microsoft/gctoolkit/sample/Main.java
```diff
@@ -7,20 +7,26 @@
 import com.microsoft.gctoolkit.sample.aggregation.HeapOccupancyAfterCollectionAggregation;
 import com.microsoft.gctoolkit.sample.aggregation.HeapOccupancyAfterCollectionSummary;
 
+import java.nio.file.Files;
```
dsgrieve, line 10: Should remove this import.

sample/src/main/java/com/microsoft/gctoolkit/sample/Main.java
```diff
@@ -7,20 +7,26 @@
...
-        GCLogFile logFile = new SingleGCLogFile(Path.of("./gclogs/preunified/cms/defnew/details/defnew.log"));
+    public void execute(String gcLogFilePath) {
+        GCLogFile logFile = new SingleGCLogFile(Path.of(gcLogFilePath));
         GCToolKit gcToolKit = new GCToolKit();
         JavaVirtualMachine machine = gcToolKit.analyze(logFile);
-        HeapOccupancyAfterCollectionAggregation results = machine.getAggregation(HeapOccupancyAfterCollectionSummary.class);
-        System.out.println(results.toString());
+        HeapOccupancyAfterCollectionSummary results = machine.getAggregation(HeapOccupancyAfterCollectionSummary.class);
+        results.get().forEach((gcType, dataSet) ->
+                System.out.println("The XYDataSet for " + gcType + " contains " + dataSet.size() + " items.")
```
dsgrieve, line 29: I think this is more interesting output, but it does not match the output in the introductory blog post.

sample/src/main/java/com/microsoft/gctoolkit/sample/Main.java
dsgrieve: Also... we should remove the `clean.sh` script from the top-level.

sample/src/main/java/com/microsoft/gctoolkit/sample/Main.java
dsgrieve: Done in commit 5a8f2ee

dsgrieve (approved):

PR #34, opened 2021-08-11
remove dependency on top-level gctoolkit-testdata

dsgrieve:
The top-level POM file has a dependency on gctoolkit-testdata. This means that when someone runs a build, the gctoolkit-testdata pom file is downloaded. But nothing is done with that dependency. The individual gclog zip files are still downloaded by the maven-dependecy-plugin. 

The other benefit of doing this is that someone can run a maven build with -DskipTests -DskipUnpack and the gctoolkit-testdata will not be downloaded. This could be recommended for someone who wants to build but doesn't want to test. It is unfortunate that skipUnpack cannot be conditional on skipTests. But we can solve that by defaulting skipUnpack and skipTests to true, and having a `contributor` profile that sets them to false. I'll create an issue for that. 

Oh, and I doc'd skipUnpack in README

### DIFF
--- a/README.md
+++ b/README.md
@@ -34,8 +34,6 @@ You must also add `github` as a server in your `~/.m2/settings.xml` file. Replac
     </server>
 ```
 
----
-
 ## Getting Started
 
 ### Maven Coordinates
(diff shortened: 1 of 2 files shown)


karianna (approved):
LGTM

PR #39, opened 2021-08-12
Added comparable interface for DateTimeStamp and its implementation.

anantk24:
- Added null last comparator
- changed access modifiers according to JLS.

### DIFF
--- a/core/api/src/main/java/com/microsoft/gctoolkit/time/DateTimeStamp.java
+++ b/core/api/src/main/java/com/microsoft/gctoolkit/time/DateTimeStamp.java
@@ -5,7 +5,10 @@ package com.microsoft.gctoolkit.time;
 import java.time.ZonedDateTime;
 import java.time.format.DateTimeFormatter;
 import java.time.temporal.TemporalAccessor;
+import java.util.Comparator;
 import java.util.Objects;
+
+import static java.util.Comparator.*;
 
 /**
  * A date and time. Both date and time come from reading the GC log. In cases where
@@ -15,7 +18,7 @@ import java.util.Objects;
  * Instance of DateTimeStamp are created by the parser. The constructors match what might be
  * found for dates and time stamps in a GC log file.
  */
-public class DateTimeStamp {
+public class DateTimeStamp implements Comparable<DateTimeStamp> {
     // Represents the time from Epoch
     // In the case where we have timestamps, the epoch is start of JVM
     // In the case where we only have date stamps, the epoch is 1970:01:01:00:00:00.000::UTC+0
@@ -27,11 +30,12 @@ public class DateTimeStamp {
     // Requirements
     // Timestamp can never be less than 0
     //      - use NaN to say it's not set
-    final private ZonedDateTime dateTime;
-    final private double timeStamp;
+    private final ZonedDateTime dateTime;
+    private final double timeStamp;
+    public static final Comparator<DateTimeStamp> comparator = getComparator();
 
     // For some reason, ISO_DATE_TIME doesn't like that time-zone is -0100. It wants -01:00.
-    private final static DateTimeFormatter formatter = DateTimeFormatter.ofPattern("yyyy-MM-dd'T'HH:mm:ss.SSSZ");
+    private static final DateTimeFormatter formatter = DateTimeFormatter.ofPattern("yyyy-MM-dd'T'HH:mm:ss.SSSZ");
 
     private static ZonedDateTime fromString(String iso8601DateTime) {
         if (iso8601DateTime != null) {
@@ -252,4 +256,11 @@ public class DateTimeStamp {
         return this.minus(dateTimeStamp) / 60.0d;
     }
 
+    @Override
+    public int compareTo(DateTimeStamp dateTimeStamp) {
+        return  comparator.compare(this,dateTimeStamp);
+    }
+    private static  Comparator<DateTimeStamp> getComparator(){
+        return nullsLast(comparing(DateTimeStamp::getDateTime));
+    }
 }
--- a/core/api/src/main/java/com/microsoft/gctoolkit/time/DateTimeStamp.java
+++ b/core/api/src/main/java/com/microsoft/gctoolkit/time/DateTimeStamp.java
@@ -8,7 +8,9 @@ import java.time.temporal.TemporalAccessor;
 import java.util.Comparator;
 import java.util.Objects;
 
-import static java.util.Comparator.*;
+import static java.util.Comparator.comparing;
+import static java.util.Comparator.nullsLast;
+
 
 /**
  * A date and time. Both date and time come from reading the GC log. In cases where
@@ -92,7 +94,7 @@ public class DateTimeStamp implements Comparable<DateTimeStamp> {
         this.dateTime = dateTime;
         //NaN is our agreed upon not set but less than 0 makes no sense either.
         if ( dateTime != null && (Double.isNaN(timeStamp) || timeStamp < 0.0d))
-            this.timeStamp = (double)dateTime.toEpochSecond() + (double) dateTime.getNano() / 1_000_000_000d;
+            this.timeStamp = dateTime.toEpochSecond() +  dateTime.getNano() / 1_000_000_000d;
         else
             this.timeStamp = timeStamp;
     }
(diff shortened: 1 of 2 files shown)


core/api/src/main/java/com/microsoft/gctoolkit/time/DateTimeStamp.java
```diff
@@ -5,8 +5,11 @@
 import java.time.ZonedDateTime;
 import java.time.format.DateTimeFormatter;
 import java.time.temporal.TemporalAccessor;
+import java.util.Comparator;
 import java.util.Objects;
 
+import static java.util.Comparator.*;
```
karianna, line 11: Prefer explicit imports

core/api/src/main/java/com/microsoft/gctoolkit/time/DateTimeStamp.java
```diff
@@ -252,4 +256,11 @@ public double timeSpanInMinutes(DateTimeStamp dateTimeStamp) {
         return this.minus(dateTimeStamp) / 60.0d;
     }
 
+    @Override
+    public int compareTo(DateTimeStamp dateTimeStamp) {
+        return  comparator.compare(this,dateTimeStamp);
+    }
```
karianna, line 262: Needs a newline here

core/api/src/test/java/com/microsoft/gctoolkit/test/time/DateTimeStampTest.java
```diff
@@ -353,5 +357,20 @@ void timeSpanInMinutes() {
         diff = a.timeSpanInMinutes(b);
         assertEquals((.18 - .19)/60d, diff, .001);
     }
+    @Test
+    void testCompare(){
```
karianna, line 361: Prefer tests that have one assert so perhaps test less than, equals to and greater than?

karianna (changes requested):

core/api/src/main/java/com/microsoft/gctoolkit/time/DateTimeStamp.java
```diff
@@ -252,4 +256,11 @@ public double timeSpanInMinutes(DateTimeStamp dateTimeStamp) {
...
+    private static  Comparator<DateTimeStamp> getComparator(){
+        return nullsLast(comparing(DateTimeStamp::getDateTime));
```
dsgrieve, line 264: DateTimeStamp is kind of a tricky thing because it depends on the log file flags. Sometimes, there is no ISO date/time in the log. Sometimes there isn't even a time stamp (which represents milliseconds since the start of the JVM). Indeed, most of the time we only care about the time stamp. 

The way this class is written and used is to try to ensure there is always a timeStamp. It is possible that timeStamp could be NaN, or less than zero - neither of which make sense and point to a bug in the code that calls the constructor. 

So you may or may not have a dateTime, but you should always have a timeStamp >= 0. One possible way to do this compare is to compare the dateTime (if not null) and if they are the same, then compare the timeStamp. Another possible way would be to add the timeStamp value to the dateTime (if not null) and compare that.

core/api/src/test/java/com/microsoft/gctoolkit/test/time/DateTimeStampTest.java
```diff
@@ -353,5 +353,46 @@ void timeSpanInMinutes() {
         diff = a.timeSpanInMinutes(b);
         assertEquals((.18 - .19)/60d, diff, .001);
     }
+    @Test
+    void testCompareLessThan(){
+        DateTimeStamp smaller = new DateTimeStamp("2018-04-04T09:10:00.586-0100");
+        DateTimeStamp greater = new DateTimeStamp("2018-04-04T10:10:00.587-0100");
+        assertEquals(-1, smaller.compareTo(greater));
+    }
+    @Test
```
karianna, line 362: Nitpick = Whitespace between each of the tests

core/api/src/main/java/com/microsoft/gctoolkit/time/DateTimeStamp.java
```diff
@@ -252,4 +258,23 @@ public double timeSpanInMinutes(DateTimeStamp dateTimeStamp) {
...
+     * @param dateTimeStamp - other object to compared
+     * @return  a negative integer, zero, or a positive integer as this object is less than, equal to, or greater than the specified object
+     */
+    @Override
+    public int compareTo(DateTimeStamp dateTimeStamp) {
+        return  comparator.compare(this,dateTimeStamp);
+    }
+
+    private static  Comparator<DateTimeStamp> getComparator(){
+        // compare withs dateTime if null, then it will go to last
```
karianna, line 273: Typo in the comments

karianna (changes requested):
2 small nitpicks but the rest is looking good to me!

karianna (approved):
LGTM

anantk24:
@kirk-microsoft waiting for your comments :) Kindly review .

dsgrieve (changes requested):

core/api/src/test/java/com/microsoft/gctoolkit/test/time/DateTimeStampTest.java
```diff
@@ -353,5 +353,46 @@ void timeSpanInMinutes() {
...
+    void testCompareSmallerTimeStamp(){
+        DateTimeStamp smaller = new DateTimeStamp("2018-04-04T10:10:00.586-0100",123);
+        DateTimeStamp greater = new DateTimeStamp("2018-04-04T10:10:00.586-0100",124);
+        assertEquals(-1, smaller.compareTo(greater));
+    }
+    @Test
+    void testCompareNullValue(){
+        DateTimeStamp smaller = new DateTimeStamp("2018-04-04T09:10:00.586-0100");
+        assertEquals(-1, smaller.compareTo(null));
+    }
```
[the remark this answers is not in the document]
anantk24: Seems, needs improvement in code when we have 
`new DateTimeStamp(123).compareTo(new `DateTimeStamp("2018-04-04T10:10:00.586-0100"))`
Should I assume when DateTime is null then directly compare timestamp, instead of comparing DateTime first?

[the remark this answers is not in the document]
dsgrieve: This is where I'd like @kirk-microsoft to chime in. KIrk, should the compare just be timeStamp only.

anantk24:
> Hi Anant, first thank you for taking the time to create this contribution. The question I have with this PR stems from the pattern for using comparator in the JDK. If you look at how it's used, the class will implement the interface and then provide an implementation for the method. I maybe mistaken by I don't recalling seeing static method support for comparator. I should also let you know, the coding style that we've been following is to avoid static methods as these are often needed to cover over missing design elements (we call this a code smell, you might find this short article interesting if you're unaware of the term. https://medium.com/oceanize-geeks/code-smells-and-refactoring-c2c0e0642582). Thus I always tend to like to probe a wee bit to see if that missing element is actually missing and if it is missing, then, how to we fix that and if not, then static methods it is... If you have a use case for how you'd want to use these static methods in this code base, then let's get this merged. If not, I think it's best if the implementation was trimmed down to the more traditional pattern.
> 
> Other than that, it's pretty much as David stated. In the majority of logs we'll only have the time stamp (age of JVM). In a significant number of logs we also have a date stamp. On occasion we only have the date stamp and part of the work performed is to estimate the age of the JVM. On rare occasions, we don't see either at which point.. well, it's a mess and in this code base, I'm not exactly sure what would happen. In the older (pre-module) code base, it would plot everything at time 0...

@kirk-microsoft Thanks for comment. I agree with your point on static api, My thought behind making comparator variable static, so that the single comparator can be used to sort collection of DateTimeStamp objects along with null objects. If there will be no use case of  separate collection DateTimeStamp objects sorting, I will update code without static apis.

anantk24:
> > Hi Anant, first thank you for taking the time to create this contribution. The question I have with this PR stems from the pattern for using comparator in the JDK. If you look at how it's used, the class will implement the interface and then provide an implementation for the method. I maybe mistaken by I don't recalling seeing static method support for comparator. I should also let you know, the coding style that we've been following is to avoid static methods as these are often needed to cover over missing design elements (we call this a code smell, you might find this short article interesting if you're unaware of the term. https://medium.com/oceanize-geeks/code-smells-and-refactoring-c2c0e0642582). Thus I always tend to like to probe a wee bit to see if that missing element is actually missing and if it is missing, then, how to we fix that and if not, then static methods it is... If you have a use case for how you'd want to use these static methods in this code base, then let's get this merged. If not, I think it's best if the implementation was trimmed down to the more traditional pattern.
> > Other than that, it's pretty much as David stated. In the majority of logs we'll only have the time stamp (age of JVM). In a significant number of logs we also have a date stamp. On occasion we only have the date stamp and part of the work performed is to estimate the age of the JVM. On rare occasions, we don't see either at which point.. well, it's a mess and in this code base, I'm not exactly sure what would happen. In the older (pre-module) code base, it would plot everything at time 0...
> 
> @kirk-microsoft Thanks for comment. I agree with your point on static api, My thought behind making comparator variable static, so that the single comparator can be used to sort collection of DateTimeStamp objects along with null objects. If there will be no use case of separate collection DateTimeStamp objects sorting, I will update code without static apis.

@karianna @dsgrieve 
Any Comments?

dsgrieve (approved):

dsgrieve:
Thanks @anantk24! It is surprising at times how a seemingly simple thing can get so complicated. Personally, I learned a thing or two about Comparator from your submission. Thanks!

anantk24:
@dsgrieve Thanks David ! Even I am learning a lot from you guys ! Many more learnings to come and pull request for you guys :)

core/api/src/test/java/com/microsoft/gctoolkit/test/time/DateTimeStampTest.java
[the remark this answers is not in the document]
kirk-microsoft: if there is a time stamp, then I think we can restrict the comparison to that. At the moment, we always generate a synthetic time stamp but I think that it would be prudent to check the date stamp in the case where the time stamp is missing.

PR #40, opened 2021-08-12
add script to run the sample

kcpeppe:
Shell script to run the sample app.


sample/sample.sh
```diff
@@ -0,0 +1,10 @@
+
```
karianna, line 1: Need to set the shell

sample/sample.sh
```diff
@@ -0,0 +1,10 @@
...
+
+export APP=target/sample-2.0.1-SNAPSHOT.jar
```
karianna, line 3: Recommend extracting versions to constants at top of script

karianna (changes requested):

dsgrieve (commented):
Why why why do we need a shell script when all you have to do is `mvn exec:exec`? Either one way or the other. Otherwise it is just confusing to the end user. And if the way to do this is a shell script, then you should provide a bat or cmd script for windows users.

kcpeppe:
It's often useful to have different ways of running an app. For example, I often use the command line for testing purposes and this script helps to form a template to work from. Just like I often build and run from the IDE.

As for cmd or bat, I don't have access to a window machine so I couldn't test either at the moment.

sample/sample.sh
```diff
@@ -0,0 +1,16 @@
...
+export JACKSON=$HOME/.m2/repository/com/fasterxml/jackson/core/jackson-core/2.11.4/jackson-core-2.11.4.jar 
+
+java -p $APP:$MODULES:$VERTX:$NETTY:$JACKSON -m gctoolkit.sample/com.microsoft.gctoolkit.sample.Main ../gclogs/preunified/cms/defnew/details/defnew.log
```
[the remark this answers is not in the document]
karianna: See #64

PR #43, opened 2021-08-16
Solved spotbugs related to GCEvent class.

anantk24:
- Defined hash code api
- added null check and class check in equals api
- Modified access modifiers according to JLS.


dsgrieve (approved):

dsgrieve:
Thanks, @anantk24. Good find!

Issue #49, opened 2021-08-18
DateTimeStamp API that takes double as an arg should handle NaN 


dsgrieve:
Double.NaN should be treated as 0.0.

I would expect, for example, that `new DateTimeStamp(0.0).add(Double.NaN).equals(new DateTimeStamp(0.0))` to be true.

anantk24:
I will take this up . Assign it to me.

anantk24:
Added pull request #50

PR #50, opened 2021-08-18
Added handling for NAN in add method of DateTimeStamp

anantk24:
- When offsetInDecimalSeconds is NAN then it is considered as 0.0d
-  #49


core/api/src/test/java/com/microsoft/gctoolkit/test/time/DateTimeStampTest.java
```diff
@@ -440,4 +440,25 @@ void testCompareSmallerTimeStampMixed() {
...
+    void testNanWithNonZero() {
+        DateTimeStamp dateTimeStamp = new DateTimeStamp("2018-04-04T10:10:00.586-0100");
+        DateTimeStamp forComparing = new DateTimeStamp("2018-04-04T10:10:00.586-0100").add(Double.NaN);
+        assertEquals(dateTimeStamp, forComparing);
+    }
+
+    @Test
+    void testNanWithNegativeAdd() {
+        DateTimeStamp dateTimeStamp = new DateTimeStamp("2018-04-04T10:09:59.586-0100");
+        DateTimeStamp forComparing = new DateTimeStamp("2018-04-04T10:10:00.586-0100").add(-1);
```
dsgrieve, line 460: shouldn't this be `minus(Double.NaN)`?

dsgrieve:
Would you mind looking at all of the DateTimeStamp API that takes a double? 
* There is a hole in the constructor where new DateTimeStamp((String)null, Double.NaN) sets timeStamp = Double.NaN 
* before/after(Double.NaN) both return false (because they compare with inequality). Double.compareTo handles NaN, but if we interpret NaN as zero, then it doesn't do the right thing. I mean, Double.compareTo(42d, Double.NaN) returns -1 where we want Double.compareTo(42d, 0.0d) (which returns 1).

anantk24:
> Would you mind looking at all of the DateTimeStamp API that takes a double?
> 
> * There is a hole in the constructor where new DateTimeStamp((String)null, Double.NaN) sets timeStamp = Double.NaN
> * before/after(Double.NaN) both return false (because they compare with inequality). Double.compareTo handles NaN, but if we interpret NaN as zero, then it doesn't do the right thing. I mean, Double.compareTo(42d, Double.NaN) returns -1 where we want Double.compareTo(42d, 0.0d) (which returns 1).

Yes sure will look for all api.

anantk24:
@dsgrieve @karianna  @pliakas  Any guess why this is failing ? I am not able to understand the root cause?
Any steps to do avoid failure?

dsgrieve:
@anantk24 It's looking for a file DateTimeStampNaN which it can't find. Is this a file you created? If so, it was never pushed.

anantk24:
> @anantk24 It's looking for a file DateTimeStampNaN which it can't find. Is this a file you created? If so, it was never pushed.

It is the branch I have created ,  trying to create pull request from the same.
`commits into microsoft:main from anantk24:DateTimeStampNAN`

dsgrieve:
It looks like the build script is trying to commit the badges to gctoolkit and not your fork, which I guess makes sense. Maybe there needs to be a different script for PRs that doesn't do the badging. Or maybe there is some config for the commit script that's missing.  IDK. New territory for me. I need to investigate.

dsgrieve:
@anantk24 - I pushed a fix for the badge commit. Please pull from gctoolkit/main

anantk24:
> It looks like the build script is trying to commit the badges to gctoolkit and not your fork, which I guess makes sense. Maybe there needs to be a different script for PRs that doesn't do the badging. Or maybe there is some config for the commit script that's missing. IDK. New territory for me. I need to investigate.

@dsgrieve  - Thanks for update. I was thinking on similar lines, opened discussion  #52 for the same. Automatic badge update is available in markdown and through supported code analysis server.

anantk24:
> @anantk24 - I pushed a fix for the badge commit. Please pull from gctoolkit/main

@dsgrieve  Thanks for changes. Pulled changes from main. It failed again . Is it working for everyone else?

dsgrieve:
I'm not sure why this worked for my PR and not yours, unless it has to do with some mojo I have that you do not. Maybe committing the badge doesn't make sense for a PR build. But the git-auto-commit-action doc implies it should work for the PR. Maybe it's the detached head state of the checkout. Maybe the workaround is to have a separate PR workflow yaml.

dsgrieve:
I just recreated the issue by checking out your pr, setting upstream to my fork, and then creating a PR from there (PR #57).

The reason it happens on your branch and not others is that you added tests, thus changing the badge, resulting in an attempt to do the commit. 

The answer, I believe, lies somewhere in the [`branch` option of the git-auto-commit-action](https://github.com/stefanzweifel/git-auto-commit-action/blob/master/action.yml).

dsgrieve (approved):

dsgrieve:
Thanks again, @anantk24!

PR #54, opened 2021-08-24
Added function getTimestampValue to handle nan value .

anantk24:
- if value is NAN for timestamp , then it is considered as zero.
- added testcases for datetime constructor and nan value for timestamp.


anantk24:
Closing this there is duplicate #50 . Created for test purpose.

PR #60, opened 2021-08-27
Three small improvements

kabutz:


core/vertx/src/main/java/com/microsoft/gctoolkit/vertx/JVMEventSource.java
```diff
@@ -32,17 +32,13 @@ public void publishSafePointLogFile(Path logFile) throws IOException {
         vertx.eventBus().publish(publicationChannel, safepointLogFile.endOfData());
     }
 
-    CountDownLatch latch = new CountDownLatch(1);
+    private final StartingGun latch = new StartingGun();
```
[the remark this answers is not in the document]
kabutz: No, this is precisely what the AbstractQueuedSynchronizer is there for. I am planning to create a course about that some time soon. I spoke about it here: https://www.youtube.com/watch?v=z6yv7QWiK0s

kabutz:
Added the JavaDocs

kabutz:
Did another small refactoring to make the code more robust

karianna (approved):
I'm personally good with all of this, nice modernization changes.

core/api/src/main/java/com/microsoft/gctoolkit/event/g1gc/G1Young.java
```diff
@@ -213,9 +215,9 @@ public void tableFixupStatistics(StatisticalSummary summary) {
         this.tableFixupStatistics = summary;
     }
 
-    private HashMap<String, Double> preEvacuateCSetPhase = new HashMap<>(3);
-    private HashMap<String, UnifiedStatisticalSummary> evacuateCSetPhase = new HashMap<>();
-    private HashMap<String, Double> postEvacuateCSetPhase = new HashMap<>();
+    private final Map<String, Double> preEvacuateCSetPhase = new ConcurrentHashMap<>(3);
```
dsgrieve, line 218: What is the thinking behind changing these (and others) from HashMap to ConcurrentHashMap when the accessors are only called from a single thread from the parser.

core/api/src/main/java/com/microsoft/gctoolkit/event/zgc/ZGCConcurrentPhases.java
```diff
@@ -2,7 +2,10 @@
 // Licensed under the MIT License.
 package com.microsoft.gctoolkit.event.zgc;
 
-public enum ZGCConcurrentPhases {
+import com.microsoft.gctoolkit.event.GarbageCollectionTypes;
+import com.microsoft.gctoolkit.event.LabelledGCEventType;
+
+public enum ZGCConcurrentPhases implements LabelledGCEventType {
```
dsgrieve, line 8: Just so you know, ZGC and Shenandoah are not fully baked.

kabutz: Indeed, am I'm also getting an exception when running the sample. Hopefully this can all be finished soon.

core/api/src/main/java/com/microsoft/gctoolkit/event/g1gc/G1Young.java
kabutz: Best practice is to avoid HashMap, even in the case where we are reasonably sure that we will only ever use one thread. The downside of the HashMap is too big, in case that some code is changed to be used in parallel. http://www.javaperformancetuning.com/news/news189.shtml

PR #65, opened 2021-08-31
Cleanup

kabutz:
Looking at the code, I noticed loads of places where fields were not marked as final, but could have been. This is confusing when maintaining, because it is not clear what fields are changed later in the code.


karianna:
@kabutz - Meta point, does this supersede your other PR? I think I see the same commits in this one?

kabutz:
I was trying to create this PR in a different branch to keep things clean, because it seems to take quite a while to get PRs approved in the Microsoft world.

karianna:
> I was trying to create this PR in a different branch to keep things clean because it seems to take quite a while to get PRs approved in the Microsoft world.

We're usually a little faster, but we're down a few folks in these past weeks.

kabutz:
I see my build failed - it builds on my machine ;-) I think what happened is a timing issue in my test. I will have to fix that but will wait until the other commits have been accepted (or rejected)

karianna:
@kabutz Looks like one test failed, not sure if that's local to you as well?

karianna:
@kabutz - You should be able to rebase this now

kabutz:
Done the rebase, with a force push

karianna (approved):
LGTM - personal opinion is I like explicit finals, let's see what the other maintainers think :-)

kabutz:
I don't like finals on parameters and local variables, but for fields they are very sensible

dsgrieve:
> I don't like finals on parameters and local variables, but for fields they are very sensible

What's the harm? 

I'm a big fan of final for local variables. Using final makes sure I don't accidentally repurpose the variable. It makes you stop and think whether assigning the variable a new value is _really_ what you want to do. It says something about the variable - "Hey, this HttpConnection... it's the same one throughout this method call."

Using final makes sure the variable is initialized, avoiding coding mistakes
```java
final String foo;
if ( bar() ) foo = "bar";
...
System.out.println(foo); // foo might not have been initialized!
```

Also, in some cases, final can make the code a little faster. 
```java
String key = "foo";
Bar bar = map.get(key);
```
performs a little slower than
```java
final String key = "foo";
Bar bar = map.get(key);
```

dsgrieve (approved):

kabutz:
> > I don't like finals on parameters and local variables, but for fields they are very sensible
> 
> What's the harm?
> 
> I'm a big fan of final for local variables. Using final makes sure I don't accidentally repurpose the variable. It makes you stop and think whether assigning the variable a new value is _really_ what you want to do. It says something about the variable - "Hey, this HttpConnection... it's the same one throughout this method call."

This is not how the JDK is coded. Final local variables and parameters are very seldomly used. They were introduced in Java 1.1 to allow nested classes to access outer variables safely. However, since Java 8 we have effectively final, thus we do not need it anymore. Final local variables and parameters add unnecessary clutter.

> Using final makes sure the variable is initialized, avoiding coding mistakes
> 
> ```java
> final String foo;
> if ( bar() ) foo = "bar";
> ...
> System.out.println(foo); // foo might not have been initialized!
> ```

Nope, for local variables this is already checked. No need to make it final - that does not do anything.

> Also, in some cases, final can make the code a little faster.
> 
> ```java
> String key = "foo";
> Bar bar = map.get(key);
> ```
> 
> performs a little slower than
> 
> ```java
> final String key = "foo";
> Bar bar = map.get(key);
> ```

Can you please prove this by means of a meaningful benchmark?

dsgrieve:
Thanks for the education, @kabutz! 

Here is a benchmark, warts and all. 

```shell
jjava.version: 11.0.10
os.arch: x86_64
lookup_1 nanos: 49938695636
lookup_2 nanos: 52750825502
delta: 2812129866
```

```java
import java.util.HashMap;
import java.util.Map;
import java.util.Random;
import java.util.function.BiFunction;
import java.util.stream.IntStream;

class Scratch {

    static Map<String, Integer> map = new HashMap<>();
    static int[] keys = new int[Short.MAX_VALUE];

    static {
        IntStream.range(0, Short.MAX_VALUE).forEach(n -> map.put(makeKey(n), n));
        Random random = new Random();
        IntStream.range(0, keys.length).forEach(n -> keys[n] = random.nextInt(Short.MAX_VALUE));
    }

    static String makeKey(int n) {
        return Integer.toHexString(n);
    }

    static BiFunction<Map<?,?>, Integer, Object> lookup_1 = (m, n) -> {
        final String key = makeKey(n);
        return map.get(key);
    };

    static BiFunction<Map<?,?>, Integer, Object> lookup_2 = (m, n) -> {
        String key = makeKey(n);
        return map.get(key);
    };

    static long test(BiFunction<Map<?,?>, Integer, Object> lookup) {
        // warmup
        for (int n=0; n<3; n++) IntStream.range(0, keys.length).forEach(k -> lookup.apply(map, k));

        long t0 = System.nanoTime();
        IntStream.range(0, Integer.MAX_VALUE).forEach(k -> lookup.apply(map, k % Short.MAX_VALUE));
        long t1 = System.nanoTime();
        return t1 - t0;

    }

    public static void main(String[] args) {
        System.out.println("java.version: " + System.getProperty("java.version"));
        System.out.println("os.arch: " + System.getProperty("os.arch"));
        long nanos1 = test(lookup_1);
        long nanos2 = test(lookup_2);
        System.out.println("lookup_1 nanos: " + nanos1);
        System.out.println("lookup_2 nanos: " + nanos2);
        long delta = (nanos1 > nanos2) ? nanos1 - nanos2 : nanos2 - nanos1;
        System.out.println("delta: " + delta);
    }

}
```

karianna:
As an aside, I'd recommend using JMH to eliminate any JVM noise in this
bench.

Cheers,
Martijn

kabutz:
@kirk-microsoft has some excellent education on microbenchmarks, but in this case it would have been easier to simply disassembled the lambda code and see that the two methods are identical:

```
  private static java.lang.Object lambda$static$3(java.util.Map, java.lang.Integer);
    Code:
       0: aload_1
       1: invokevirtual #95                 // Method java/lang/Integer.intValue:()I
       4: invokestatic  #99                 // Method makeKey:(I)Ljava/lang/String;
       7: astore_2
       8: getstatic     #79                 // Field map:Ljava/util/Map;
      11: aload_2
      12: invokeinterface #102,  2          // InterfaceMethod java/util/Map.get:(Ljava/lang/Object;)Ljava/lang/Object;
      17: areturn

```

```
  private static java.lang.Object lambda$static$2(java.util.Map, java.lang.Integer);
    Code:
       0: aload_1
       1: invokevirtual #95                 // Method java/lang/Integer.intValue:()I
       4: invokestatic  #99                 // Method makeKey:(I)Ljava/lang/String;
       7: astore_2
       8: getstatic     #79                 // Field map:Ljava/util/Map;
      11: aload_2
      12: invokeinterface #102,  2          // InterfaceMethod java/util/Map.get:(Ljava/lang/Object;)Ljava/lang/Object;
      17: areturn
```

dsgrieve:
I did this also as an afterthought, with the same result. Interestingly, when I did this in the past it came up with different bytecode. Maybe the test code I used was different, or maybe it was a different java version.

PR #70, opened 2021-09-01
Removes auto loading aggregators. Requires API call

brunoborges:
I spent quite some time trying to understand how the sample application was loading the aggregator, and realized that the auto-loading feature, while nice, is too magic.

This PR suggests that the loading can still happen but requires library consuming developer to confirm this desire, by calling `loadAggregationsFromServiceLoader()` in GCToolKit.

This also enhances the `getAggregator` to return an `Optional<T>` as the aggregator may not be registered/missing.


core/api/src/main/java/com/microsoft/gctoolkit/GCToolKit.java
```diff
@@ -60,17 +51,40 @@ private static JavaVirtualMachine loadJavaVirtualMachine() {
...
+    public void loadAggregationsFromServiceLoader() {
+        if (aggregationsLoadedFromServiceLoader == true) return;
+
+        ServiceLoader.load(Aggregation.class)
+                .stream()
+                .map(ServiceLoader.Provider::get)
+                .map(Aggregation::getClass)
+                .forEach(registeredAggregations::add);
+
+        aggregationsLoadedFromServiceLoader = true;
```
tbee, line 81: Is it possible that two threads run through this block at the same time? Then classes may be registered twice. so the whole block including boolean check should have a lock around it. The fact that it was in static {} before, made sure it was only executed once per loaded class.

brunoborges: I believe my mistake here was letting the boolean as a static field. It shouldnt be. Would that clear your concern?

kcpeppe:
I'm not sure why we're removing this feature. To not have autoloading affects the ability to clients to load queries without modifying code. Can we put this into an issue where we can discuss it further

core/api/src/main/java/com/microsoft/gctoolkit/GCToolKit.java
```diff
@@ -60,17 +51,40 @@ private static JavaVirtualMachine loadJavaVirtualMachine() {
...
+     * import com.microsoft.gctoolkit.sample.aggregation.HeapOccupancyAfterCollectionSummary;
+     * 
+     * module com.microsoft.gctoolkit.sample {
+     *     ...
+     *     provides Aggregation with HeapOccupancyAfterCollectionSummary;
+     * }
+     * </pre>
+     */
+    public void loadAggregationsFromServiceLoader() {
+        if (aggregationsLoadedFromServiceLoader == true) return;
```
dsgrieve, line 73: I don't think this flag/check is necessary. Someone may have reason to call this more than once (dynamically load new aggregations). Unlikely, but possible. Besides, `registeredAggregators` is a set, so you will not end up with more than one entry for a given aggregation class.

core/api/src/main/java/com/microsoft/gctoolkit/jvm/JavaVirtualMachine.java
```diff
@@ -91,7 +93,8 @@
      * {@code Aggregation}s are used depends on the GC.
      * @param aggregationClass The class of the Aggregation to be returned.
      * @param <T> type cast for the Aggregation class type.
-     * @return an {@code Aggregation} whose {@code getClass() == aggregationClass}, or {@code null}
+     * @return an {@code Aggregation} whose {@code getClass() == aggregationClass}, or {@code null} 
+     * if given aggregationClass is not available.
      */
-    <T extends Aggregation> T getAggregation(Class<T> aggregationClass);
+    <T extends Aggregation> Optional<T> getAggregation(Class<T> aggregationClass);
```
dsgrieve, line 99: Maybe. I think Optional only really helps if you are using this API in a stream. Do you have a use-case/example for this?

dsgrieve:
For discussion: 
---
Bruno and I had a discussion about this API in general. It was very refreshing to have his un-initiated perspective here. One of the ideas he tossed into the ring was having an 'analyzer' API, to which my response was that is what GCToolkit is. 

Later, I got to thinking about it and I see where he's coming from. I think, perhaps, we should rename the GCToolKit class to `GCAnalyzer`, and have the `analyze` method return a `GCAnalysis` which would be what JavaVirtualMachine is now. I think this would make it more clear to the user what to do and what to use. 

Thoughts?

dsgrieve (commented):
> I'm not sure why we're removing this feature. To not have autoloading affects the ability to clients to load queries without modifying code. Can we put this into an issue where we can discuss it further

This does not remove the feature, but puts it front-and-center rather than being an obscured piece of magic (for those who don't rtfm).

core/api/src/main/java/com/microsoft/gctoolkit/jvm/JavaVirtualMachine.java
brunoborges: Whenever the return value may be null, I always prefer to see an Optional<T> so I know it may be null without having to even read the javadoc.

core/api/src/main/java/com/microsoft/gctoolkit/GCToolKit.java
brunoborges: Interesting thought. Agree it makes sense to simply not have it. Besides, it's not like someone would intentionally call this more than once *unless* they have strong reasons to, as you said. Otherwise it would be a bug in their code.

brunoborges: Fixed in 4618237

core/api/src/main/java/com/microsoft/gctoolkit/GCToolKit.java
brunoborges: Fixed in 4618237

tbee: My concern was the fact that between the check and the assignment a context swap could be possible and a second threat could be waltzing in and execute the same logic also. The check-assignment approach is not a good lock. But since it is gone now, so is the concern.

core/api/src/main/java/com/microsoft/gctoolkit/jvm/JavaVirtualMachine.java
brunoborges: It's either this or adding a method `hasAggregation(Class<T>)`.

kcpeppe:
It sounds good but we’re not offering a GC analysis.

core/api/src/main/java/com/microsoft/gctoolkit/jvm/JavaVirtualMachine.java
dsgrieve: Fair enough.

dsgrieve (approved):

core/api/src/main/java/com/microsoft/gctoolkit/jvm/JavaVirtualMachine.java
kirk-microsoft: I like the enthusiasm but this thread is now mixing a couple of different concerns that is complicating the ability to discuss each of them. On point 1, one of the issues with Censum is that adding queries violated Open-Closed. Using the module version of the SPI solves this issue. To undo this features moves us back to having to violate Open-Closed. Again, SPI isn't magic, it's a well known pattern for providing implementations for an API.
On point 2, I'd like more discussion before we start renaming things. There is a model in play and I think that we should take more time to understand it before changing it.

kirk-microsoft (approved):

kirk-microsoft (approved):

PR #76, opened 2021-09-01
Unit test to demonstrate that the DateTimeStamp compareTo function is…

kabutz:
… faulty - it should be transitive

The compareTo function in DataTimeStamp has a bug. It is not transitive, but it should be. I have not fixed it, because I'm not sure how you want to do the comparison. However, I have written a unit test to demonstrate that it is broken at the moment.


karianna:
@kabutz - Murphy's law on that test failing :-)

PR #84, opened 2021-09-02
Force `.` as decimal separator DateTimeStamp's string representation

sormuras:
This commit fixes a failing test, when the default locale uses a different decimal separator character, e.g. `,` in German locale.


sormuras:
Dooh, too late and didn't sync my fork... never mind.

Although, `Locale.ROOT` is "is more neutral" and "is also safe against possible future changes in a country's choice of decimal separator." ... source: https://stackoverflow.com/a/5236096/1431016

PR #90, opened 2021-09-09
VS Code DevContainer Config and TestData Download with Maven

brunoborges:


contributor.sh
```diff
@@ -0,0 +1,109 @@
...
+revert() {
+    if [ -f .mvn/maven.config ]; then
+        rm .mvn/maven.config
+    fi
+    echo "./mvn/maven.config file removed."
+}
+
+# Download latest GCToolkit Test Data pack
+download() {
+    rm -rf gctoolkit-testdata
```
karianna, line 30: You always want to clean?

contributor.sh
```diff
@@ -0,0 +1,109 @@
...
+
+    rm -rf gctoolkit-testdata
+
+    echo "GCToolKit Test Data downloaded and installed in your local Maven repository."
+}
+
+# Prints usage help
+printhelp() {
+    cat << EOF
+Microsoft GCToolKit Test Data
```
karianna, line 53: Maybe GCToolKit Test Data Installer?

karianna: Or GCToolKit Contributor Setup?

contributor.sh
```diff
@@ -0,0 +1,109 @@
...
+-r: reverts the .mvn/maven.config file to disable 'contributor' Maven profile.
+EOF
+    exit 0
+}
+
+# MAIN
+
+# Parse flags
+
+if [ "$#" -eq 0 ]; then printhelp; fi
```
karianna, line 69: Some of this logic could be in functions...

karianna (changes requested):

contributor.sh
brunoborges: Pretty much, given this is a temporary folder. I will actually rename it with a period to hide in *nix systems.

contributor.sh
brunoborges: Installer. Installer is good.

contributor.sh
brunoborges: I guess what is missing here is an 'exit 0'

brunoborges:
This PR now significantly improves the experience with GCToolKit project by not relying on `gctoolkit-testdata` artifacts hosted on the GitHub Package Registry.

Instead, it will use Maven/Ant to download the tag (currently, `1.0.2`) source that is in GitHub Releases for the gctoolkit-testdata project, extract, and construct the `gclogs` folder structure.

The `pom.xml` configuration remains the same, meaning `contributor` profile is not enabled by default, and the download of test data, plus tests, will only happen when `contributor` profile is enabled.

I've also kept the `contributor.sh` shell script with an enable/disable feature for contributor profile, and have made the DevContainer configuration (for VS Code on Codespaces), to enable `contributor` profile by default.

karianna (approved):

karianna:
Could be unrelated, but you have a test failure.

dsgrieve (changes requested):
The test fails with this PR, but not on the main branch. I wonder how the uber-zip was created. Are you aware that the gctoolkit-testdata repo uses github LFS? Could that be in play here?

dsgrieve:
I believe that LFS is the issue. 
From the uber-zip:
133 Sep 13 09:46 gclogs/preunified/ps/details/tenuring/p_app_02_gc.log.zip

From https://maven.pkg.github.com/microsoft/*/com/microsoft/gctoolkit/gctoolkit-gclogs/1.0.2
55298582 Aug  6 16:59 gclogs/preunified/ps/details/tenuring/p_app_02_gc.log.zip

pom.xml
```diff
@@ -285,49 +297,49 @@
...
-                                    <artifactId>gctoolkit-zgc-logs</artifactId>
-                                    <version>${gctoolkit-testdata-version}</version>
-                                    <type>zip</type>
-                                </artifactItem>
-                            </artifactItems>
-                            <includes>**/*</includes>
-                            <outputDirectory>${project.basedir}/gclogs
-                            </outputDirectory>
+                            <target>
+                                <get skipexisting="true" src="https://github.com/microsoft/gctoolkit-testdata/archive/refs/tags/${gctoolkit-testdata-version}.zip" dest="${maven-parent.basedir}"/>
```
dsgrieve, line 331: This is the src zip file from the release. The source uses LFS and since src.zip is just a simple zip of the source, the large file links are not resolved. Need to look into creating a release where src.zip has LFS links resolved.

dsgrieve: Seems relevant: https://stackoverflow.com/questions/65110940/zipping-git-lfs-binaries-in-github-actions-and-adding-them-as-asset-to-a-release

brunoborges: Changes are in the main branch on my fork. PR coming today.

https://github.com/brunoborges/gctoolkit-testdata/blob/main/.github/workflows/release.yml

PR #92, opened 2021-09-16
suppress unchecked cast warnings

dsgrieve:
Suppress annoying compiler warning. The uncheck casts in JVMEventDispatcher are safe. The one in MRUQueue could still result in a ClassCastException, but we know from how/where MRUQueue is used that a ClassCastException will not occur.


karianna (approved):
LGTM

Issue #93, opened 2021-09-16
Remark is lost when ScavangeBeforeRemark is set


kcpeppe:
The Remark event is lost when ScavengeBeforeRemark is set. Test with tests will be submitted when the issues with github-teestdata are resolved.

PR #94, opened 2021-09-16
fix issue #93 with UnifiedGenerationalHeapParser

kcpeppe:
The Remark event was being lost by the UnifiedGenerationalHeapParser when ScavengeBeforeRemark was set. The remark is now held in a separate forward reference and events are queued to retain proper time order.

There is a corresponding test log file with tests that will be submitted when the test-data project has been rearranged.


karianna:
@kcpeppe You need to rebase :-)

PR #95, opened 2021-09-17
add RuntimeAggregation

dsgrieve:
Add RuntimeAggregation, which accumulates time of first event, time of last event, and runtime duration. This class is useful as a base for aggregations that need runtime to perform calculations. For example, to calculate the ratio of pause time to runtime.

### DIFF
--- a/core/api/src/main/java/com/microsoft/gctoolkit/aggregations/RuntimeAggregation.java
+++ b/core/api/src/main/java/com/microsoft/gctoolkit/aggregations/RuntimeAggregation.java
@@ -1,0 +1,75 @@
+package com.microsoft.gctoolkit.aggregations;
+
+import com.microsoft.gctoolkit.aggregator.Aggregation;
+import com.microsoft.gctoolkit.time.DateTimeStamp;
+
+/**
+ * An Aggregation that collates runtime data. This class is meant to be extended by other
+ * implementations that need access to runtime data to perform calculations. For example,
+ * to calculate the ratio of pause time to runtime duration.
+ */
+public abstract class RuntimeAggregation  implements Aggregation {
+
+    private volatile DateTimeStamp timeOfFirstEvent = null;
+    private volatile DateTimeStamp timeOfLastEvent = new DateTimeStamp(0d);
+
+    /**
+     * Runtime duration (in decimal seconds) after which we no longer consider the GC log to be a fragment
+     */
+    private static final double LOG_FRAGMENT_THRESHOLD = 18d;
+
+    /**
+     * This class is meant to be extended.
+     */
+    protected RuntimeAggregation() {}
+
+    /**
+     * RuntimeAggregation collates the time of an event and the duration of the event.
+     * @param eventTime The time a JVMEvent occurred.
+     * @param duration The duration of the JVMEvent.
+     */
+    public void record(DateTimeStamp eventTime, double duration) {
+        duration = !Double.isNaN(duration) ? duration : 0d;
+
+        if (timeOfFirstEvent == null || (eventTime != null && eventTime.before(timeOfFirstEvent))) {
+            timeOfFirstEvent = eventTime != null ? eventTime : new DateTimeStamp(0d);
+        }
+
+        final DateTimeStamp now =
+                eventTime != null ? eventTime.add(duration) : timeOfLastEvent.add(duration);
+        if (now.after(timeOfLastEvent)) {
+            timeOfLastEvent = now;
+        }
+    }
+
+    /**
+     * Return the time of the first event of the GC log.
+     * @return The time of the first event.
+     */
+    public DateTimeStamp getTimeOfFirstEvent() {
+        return timeOfFirstEvent != null ? timeOfFirstEvent : new DateTimeStamp(0.0);
+    }
+
+    /**
+     * Return the time of the last event of the GC log.
+     * Note well! The time of the last event is not the start time of the event, but is the
+     * time the event ended (event start time plus the event duration).
+     * @return The time of the last event.
+     */
+    public DateTimeStamp getTimeOfLastEvent() {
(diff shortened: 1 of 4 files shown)


core/aggregations/src/main/java/com/microsoft/gctoolkit/aggregations/RuntimeAggregation.java
```diff
@@ -0,0 +1,75 @@
...
+
+    /**
+     * RuntimeAggregation collates the time of an event and the duration of the event.
+     * @param eventTime The time a JVMEvent occurred.
+     * @param duration The duration of the JVMEvent.
+     */
+    public void record(DateTimeStamp eventTime, double duration) {
+        duration = !Double.isNaN(duration) ? duration : 0d;
+
+        if (timeOfFirstEvent == null || (eventTime != null && eventTime.before(timeOfFirstEvent))) {
```
kirk-microsoft, line 34: This will fail in the case of clock rollbacks. I would suggest that the first time a clock is seen, that is the start time of the first JVMEvent

core/aggregations/src/main/java/com/microsoft/gctoolkit/aggregations/RuntimeAggregation.java
```diff
@@ -0,0 +1,75 @@
...
+            timeOfFirstEvent = eventTime != null ? eventTime : new DateTimeStamp(0d);
+        }
+
+        final DateTimeStamp now =
+                eventTime != null ? eventTime.add(duration) : timeOfLastEvent.add(duration);
+        if (now.after(timeOfLastEvent)) {
+            timeOfLastEvent = now;
```
kirk-microsoft, line 41: In the case of clock rollbacks in the log, it's best to let the timeOfLastEvent take on the new value. This will likely result in a corrupted calculation, maybe 2.. but after that if there are no rollbacks, then all off the subsequent calculations of now should be ok

core/aggregations/src/main/java/com/microsoft/gctoolkit/aggregations/RuntimeAggregation.java
```diff
@@ -0,0 +1,75 @@
...
+    }
+
+    /**
+     * Return the duration of the GC log. Fundamentally, this is the difference between the
+     * time of the last event and the time of the first event.
+     * @return The time of the first event.
+     */
+    public double getRuntimeDuration() {
+        double duration = getTimeOfLastEvent().minus(getTimeOfFirstEvent());
+        boolean isLogFragment = duration < LOG_FRAGMENT_THRESHOLD;
```
kirk-microsoft, line 70: I think the duration should be the duration and that it being a log fragment is a different issue.

core/aggregations/src/main/java/com/microsoft/gctoolkit/aggregators/RuntimeAggregator.java
```diff
@@ -0,0 +1,31 @@
...
+/**
+ * An Aggregator that collects only the DateTimeStamp and duration of a JVMEvent.
+ */
+public class RuntimeAggregator extends Aggregator<RuntimeAggregation> {
+
+    protected RuntimeAggregator(RuntimeAggregation aggregation) {
+        super(aggregation);
+    }
+
+    // This is atypical of an Aggregator. The typical pattern is to register the method that processes
```
kirk-microsoft, line 17: I'm trying to understand why this code if following a different pattern than the norm established by registering the callback.

core/aggregations/src/main/java/com/microsoft/gctoolkit/aggregations/RuntimeAggregation.java
dsgrieve: This is copied from other code. If it is wrong here then it is wrong there. But I agree that this kind of check doesn't belong here. It should be up to the caller to decide whether or not the duration is significant.

core/aggregations/src/main/java/com/microsoft/gctoolkit/aggregations/RuntimeAggregation.java
dsgrieve: I'll look into this.

core/aggregations/src/main/java/com/microsoft/gctoolkit/aggregations/RuntimeAggregation.java
kirk-microsoft: yes, this appears to be an over-reach

karianna:
@dsgrieve Will need a rebase and then let's see if we can get this landed (Assuming you and @kirk-microsoft are now aligned)

core/aggregations/src/main/java/com/microsoft/gctoolkit/aggregations/RuntimeAggregation.java
dsgrieve: In Censum, timeOfLastEvent is calculated as given.
```
            vertx.eventBus().
                    consumer(mailBox, message -> {
                        try {
                            JVMEvent event = (JVMEvent) message.body();
                            if (event instanceof JVMTermination)
                                return;
                            DateTimeStamp now = event.getDateTimeStamp().add(event.getDuration());
                            if (now.after(timeOfLastEvent)) {
                                timeOfLastEvent = now;
                            }
                        } catch (Throwable t) {
                            LOGGER.throwing(this.getClass().getName(), "start", t);
                        }
                    });

```

core/aggregations/src/main/java/com/microsoft/gctoolkit/aggregations/RuntimeAggregation.java
dsgrieve: Fixed

karianna:
@kirk-microsoft Can you submit an official review through the GH interface?  I won't merge this just yet

karianna (approved):
LGTM

PR #97, opened 2021-09-23
Xy data set refactor

kabutz:


sample/pom.xml
```diff
@@ -24,7 +24,7 @@
             <groupId>com.microsoft.gctoolkit</groupId>
             <artifactId>gctoolkit-vertx</artifactId>
             <version>${project.version}</version>
-        </dependency>
+        </dependency><dependency><groupId>org.junit.jupiter</groupId><artifactId>junit-jupiter-api</artifactId></dependency><dependency><groupId>org.junit.jupiter</groupId><artifactId>junit-jupiter-engine</artifactId></dependency>
```
karianna, line 27: Looks like whitespace got mangled here

karianna (changes requested):
One small nitpick

sample/pom.xml
kabutz: I'll undo that - sorry 'bout that

kabutz: IntelliJ added that dependency when I created the test package in the samples module.

karianna (approved):

karianna:
Over to @kirk-microsoft - Is this the right place to move the set to?

karianna:
@kabutz - Apologies but some other code landed in the meantime, are you able to rebase?

kabutz:
Rebase done

sample/src/main/java/com/microsoft/gctoolkit/sample/collections/XYDataSet.java
```diff
@@ -1,24 +1,21 @@
 // Copyright (c) Microsoft Corporation.
 // Licensed under the MIT License.
-package com.microsoft.gctoolkit.collections;
+package com.microsoft.gctoolkit.sample.collections;
```
kirk-microsoft, line 3: Am I to understand from this that this PR would move XYDataSet to the sample module? As it stands, XYDataSet is a class that is core to the API.

sample/src/main/java/com/microsoft/gctoolkit/sample/collections/XYDataSet.java
```diff
@@ -49,8 +49,15 @@ public XYDataSet scaleSeries(double scaleFactor) {
         return scaled;
     }
 
-    public double max() {
-        return dataSeries.stream().map(Point::getY).mapToDouble(Number::doubleValue).max().getAsDouble();
+    /**
+     * Returns the largest Y value in the XYDataSet as an OptionalDouble,
+     * with an empty optional if the dataset is empty.
+     */
+    public OptionalDouble maxOfY() {
```
kirk-microsoft, line 56: curious to know why the move to Optional?

sample/src/main/java/com/microsoft/gctoolkit/sample/collections/XYDataSet.java
kabutz: It certainly does not seem to have anything to do with core, as it would be for displaying datasets. IMHO it would be better out of core.

sample/src/main/java/com/microsoft/gctoolkit/sample/collections/XYDataSet.java
kabutz: The previous code would throw an exception if you run it with an empty XYDataSet. In other words, new XYDataSet().max() throws an exception.

PR #99, opened 2021-09-23
Sorted the modifiers of fields according to best practices

kabutz:
"If two or more (distinct) field modifiers appear in a field declaration, it is customary, though not required, that they appear in the order consistent with that shown above in the production for FieldModifier."

In other words:

public protected private static final transient volatile

https://docs.oracle.com/javase/specs/jls/se17/html/jls-8.html#jls-8.3.1


karianna (approved):
LGTM

PR #101, opened 2021-09-24
New project structure

brunoborges:


sample/src/test/java/com/microsoft/gctoolkit/sample/test/TestMain.java
```diff
@@ -0,0 +1,15 @@
+package com.microsoft.gctoolkit.sample.test;
+
+import org.junit.jupiter.api.Assertions;
+import org.junit.jupiter.api.Test;
+
+import com.microsoft.gctoolkit.sample.Main;
+
+public class TestMain {
+
+    @Test
```
karianna, line 10: This is not a test, this is a runner.

karianna: Use JUnit's assertNoException syntax.  I forget exactly what it is.

karianna (changes requested):

pom.xml
```diff
@@ -480,18 +437,6 @@
     </reporting>
 
     <profiles>
-        <profile>
```
karianna, line 483: Reinstate this please.

karianna: Or add the modules we do want in the release profile instead

karianna:
`./mvnw versions:display-plugin-updates` (you can update everything except for license plugin)

karianna:
`./mvnw versions:display-dependency-updates` update deps please

brunoborges:
> Or add the modules we do want in the release profile instead 

Fixed in 885359c

pom.xml
brunoborges: Fixed in 885359c

karianna (approved):

sample/src/test/java/com/microsoft/gctoolkit/sample/test/TestMain.java
brunoborges: Fixed in 22de7a7

dsgrieve (approved):

brunoborges:
Hey @kabutz, wanna provide some input here?

kabutz:
> Hey @kabutz, wanna provide some input here?

For some reason GitHub is not showing me the individual commits at the moment. However, I see that both @karianna and @dsgrieve have already approved it.

kabutz:
> > Hey @kabutz, wanna provide some input here?
> 
> For some reason GitHub is not showing me the individual commits at the moment. However, I see that both @karianna and @dsgrieve have already approved it.

Ok, managed to see them now. What I do think should happen is that we need to carefully review all the classes in the modules and see which are internal and which should be exported. Since we do not have an adequate use case of the core and parser in the project that demonstrates how it will be used, we will need input from @kirk-microsoft. I like marking all classes as final, unless we specifically want them to be extended. Similarly, I prefer only exporting those packages from the module that I want others to access. The rest should be moved to some .internal.** package. And individual classes in packages that should not be seen from outside the packages should be package-private.

Then, one thing that bothers me about the current structure is that the tests are in different packages to the classes being tested. This means that it becomes impossible to write unit tests for the package-private classes. I know this decision would have been at least partially influenced by JPMS. However, I don't feel that is a good enough reason to do it this way.

brunoborges:
I agree @kabutz.

Perhaps it is worth creating an issue where we can discuss and track this specific matter around modules.

I'll take though as this PR in particular is OK to go and then we can iterate over with API-level changes as you suggest.

kirk-microsoft:
The modularization here was a first attempt at moving things about. I fully expect that as we move forward, things will continue to be moved about. For example, there is functionality in the vertx module that clearly shouldn't be there.. it's there because the support needed for the message bus is missing from the API module.

There is a tension between 100% test coverage and 0% test coverage and meeting both extremes isn't desirable IME. We had better test coverage in the past but then the tests became barriers to refactoring. It was in our previous major refactoring that we lost a number of tests. One of the things that does need improving is test coverage but I'd first like to focus on where that matters the most and that is in the parsers. As for tests in a different package, this comes from the notion that one should be able to test from public APIs. Having tests in a different package enforces this.

As for marking classes final, again there is a tension between marking things final so that the implementations cannot be locked down by extensions and yet being able to allow for useful but unexpected things to happen. One could argue that one should compose over extend and while I do favour composition over inheritance, there are places where inheritance is clearly the right thing.

I suggest we create issues for these things, we can break them down and link them together.

PR #102, opened 2021-09-25
Pom cleanup: Added values for name,descrption,url tag

anantk24:
- added appropriate naming conventions for name tag

- added appropriate description for each module

- added parent url for url tag


karianna:
@anantk24 Apologies but a major refactor went in so you may need to rabase or just restart this one :-(

anantk24:
> @anantk24 Apologies but a major refactor went in so you may need to rabase or just restart this one :-(



> @anantk24 Apologies but a major refactor went in so you may need to rabase or just restart this one :-(

Yes. No issues, will update it.

vertx/pom.xml
```diff
@@ -10,7 +10,9 @@
     </parent>
 
     <artifactId>gctoolkit-vertx</artifactId>
-
+    <name>GCToolKit Vertx</name>
+    <description>Event module for creating discrete event from parsing log files.</description>
```
dsgrieve, line 14: Messaging backplane for connecting GC log parsers and JVM event consumers.

dsgrieve: Just to be clear, there may be more than one parser active. A line is read from a GC log and published on the message bus, and then the parsers publish JVMEvent's on the message bus (different channels than the log line, of course!) where the events are picked up by the Aggregator's.

brunoborges (approved):
I've made commits to @anantk24 's branch, and also applied @dsgrieve 's suggestions. All seems good now.

karianna (approved):
LGTM

PR #110, opened 2021-09-28
Followed best practice in using fully qualified class names for uses and provides

kabutz:
Best practice according to "Java 9 Modularity" Mak & Bakker, as well as in the OpenJDK is to use fully qualified class names in the "uses" and "provides" clauses inside module-info.java. Thus com.microsoft.gctoolkit.jvm.JavaVirtualMachine instead of using an import and then JavaVirtualMachine. Jetty does not do it this way BTW.


karianna (approved):
LGTM

PR #111, opened 2021-09-28
Add a profile to bypass the duplicate module-info.java check in VS Code

testforstephen:
Currently opening this project in VS Code will report the error **"Build path contains duplicate entry 'module-info.java' for project x"**,  this is because VS Code Java tooling only allows a Java project to have at most one JPMS module-info.java. 

To workaround it, we can exclude test module-info.java when running in VS Code and fallback to the classpath way to handle test files. Since `m2e.version` property is only effective with VS Code Java tooling, it has no influence in Maven build itself.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -494,6 +494,34 @@
                 </plugins>
             </build>
         </profile>
+        <profile>
+            <!-- Only active in VS Code Java environment, used to bypass duplicate
+    module-info.java check in VS Code. It has no influence on the Maven build itself. -->
+            <id>m2e</id>
+            <activation>
+                <property>
+                    <name>m2e.version</name>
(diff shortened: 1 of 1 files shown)


testforstephen:
cc @brunoborges

karianna (approved):
LGTM - although we should keep an issue open for when Code supports this.  I'm guessing this would solve it for Eclipse IDE as well.

kabutz:
Do we still need this, now that we have removed the module-info.java from the test package?

testforstephen:
> Do we still need this, now that we have removed the module-info.java from the test package?

This PR becomes invalid if you have removed module-info.java from the test package. I will close it.

PR #112, opened 2021-09-28
Used canonical order for elements in module-info.java

kabutz:
Sorry for all these pull requests. I'm also figuring this JPMS stuff out as we go along. It seems that the canonical order of the elements inside the module-info.java file in the OpenJDK/src is always the same:

`module some.module.name {
  requires ...
  
  requires transitive ...

  exports ... 

  exports ... to
    ...

  opens ... to 
    ...

  uses ...

  provides ... with 
    ...
}`

The order in our module-info.java files were a bit muddled, with requires sometimes appearing near the end.

Furthermore, the entries are usually alphabetically sorted.

(There is one case in the OpenJDK where requires transitive is out of order, but that was because someone changed it and didn't reorder it. I've sent a pull-request there too, but it's so insignificant I almost doubt it will be accepted.)

### DIFF
--- a/api/src/main/java/module-info.java
+++ b/api/src/main/java/module-info.java
@@ -26,6 +26,7 @@
  * @uses com.microsoft.gctoolkit.aggregator.Aggregator
  */
 module com.microsoft.gctoolkit.api {
+    requires java.logging;
 
     exports com.microsoft.gctoolkit;
     exports com.microsoft.gctoolkit.aggregator;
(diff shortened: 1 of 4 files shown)


dsgrieve (approved):

kabutz:
https://docs.oracle.com/javase/specs/jls/se17/html/jls-7.html#jls-ModuleDirective

brunoborges (approved):
LGTM

PR #113, opened 2021-09-28
Hide starting gun

kabutz:
After our reorganization, it seems that the StartingGun class is only needed inside the vertx module. I thus propose that we move it into an internal package inside our vertx module. No point having to maintain this if someone outside of our project starts using it.


karianna:
Needs a rebase but LGTM

karianna (approved):
LGTM

brunoborges (approved):

brunoborges:
I'd like to get @kirk-microsoft's input on this one before merging.

kabutz:
> I'd like to get @kirk-microsoft's input on this one before merging.

For once I think @kirk-microsoft and even @kcpeppe will agree wholeheartedly with the proposed PR ;-)

PR #114, opened 2021-09-28
Made test methods public to avoid --add-opens

kabutz:
The reason that we needed --add-opens for our tests was because our test classes and the test methods were not public. I have changed the test classes to be public and their test methods too, so that junit does not have an issue calling them and have removed the --add-opens from the dom.xml files. I think @karianna will be pleased as punch.

### DIFF
--- a/api/pom.xml
+++ b/api/pom.xml
@@ -16,20 +16,6 @@
     <description>Model APIs for aggregator, events, inputs.</description>
     <url>${project.parent.url}</url>
 
-    <build>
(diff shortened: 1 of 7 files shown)


karianna:
@kabutz Sorry PRs are flying in so can you rebase this?

karianna:
I am pleased as punch :-)

karianna (approved):
LGTM

brunoborges (approved):

PR #129, opened 2021-10-08
Comparator date time stamp

dsgrieve:
Compare time stamps first. If equal, then compare date stamp (which is a ZonedDateTime).
If either of the date stamps are null, then consider them equal. 

The date stamps are only compared if the time stamps are not equal. In practical use, the DateTimeStamps being compared will both have non-null date stamps, or will both  have null date stamps. You won't have one date stamp null and the other not because the DateTimeStamps are created by the parser from the parsed log line. The log line will have a time stamp, a date stamp, or both a date and time stamp. Therefore, just comparing if either is null is sufficient.


brunoborges:
LGTM, and given @dsgrieve addressed @karianna's comments, I'm merging.

PR #136, opened 2021-10-13
Allow Multi-release to leverage Java 17 (11 minimum)

dsgrieve:
It would be nice to be able to take advantage of some of the language features that have been added since JDK 11.

### DIFF
--- a/.github/workflows/maven.yml
+++ b/.github/workflows/maven.yml
@@ -15,7 +15,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        java: [17]
+        java: [11, 17]
 
     steps:
     - uses: actions/checkout@v2
@@ -35,3 +35,6 @@ jobs:
         
     - name: Build with Maven
       run: ./mvnw -B verify --file pom.xml
+      env:
+        JAVA_RELEASE: $${ matrix.java }}
+
--- a/.github/workflows/maven.yml
+++ b/.github/workflows/maven.yml
@@ -15,7 +15,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        java: [11, 17]
+        java: [17]
 
     steps:
     - uses: actions/checkout@v2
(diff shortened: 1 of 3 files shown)


brunoborges:
The sort of PR where builds must fail :-)

dsgrieve:
> The sort of PR where builds must fail :-)

Heh. Works on my machine. 

Meant to be the sort of PR that engenders discussion.

rokon12:
Does it mean we can refactor some code, especially replacing the old switch case with an enhanced switch case?

brunoborges:
@dsgrieve I'm talking about this: https://docs.oracle.com/en/java/javase/11/docs/specs/jar/jar.html#multi-release-jar-files

It allows us to only revert to an older version on specific .java files, not an entire project.

kirk-microsoft:
I'm really on the fence on this one. I'm personally working in JDK 16 and will likely move to 17 very shortly. So, part of me is saying, let us run the tests in JDK 17 and if it's all green, let's do it. Then, I'm also thinking about some of the developers I know who expressed interest that are stuck using their corporate supplied laptops that have just been allowed to move up to JDK 11... so, working in JDK 17... where does it leave them. Then the other part of my brain says, do we even have any of these developers in our audience. The answer to that question... I don't know and if they are not here now, would moving to JDK 17 cause them to be excluded in the future... I simply don't know. We need a @brunoborges surgery :-)

karianna:
> I'm really on the fence on this one. I'm personally working in JDK 16 and will likely move to 17 very shortly. So, part of me is saying, let us run the tests in JDK 17 and if it's all green, let's do it. Then, I'm also thinking about some of the developers I know who expressed interest that are stuck using their corporate supplied laptops that have just been allowed to move up to JDK 11... so, working in JDK 17... where does it leave them. Then the other part of my brain says, do we even have any of these developers in our audience. The answer to that question... I don't know and if they are not here now, would moving to JDK 17 cause them to be excluded in the future... I simply don't know. We need a @brunoborges surgery :-)

This point is very valid, we may well have corporate folks (or others) who can't develop on 17.  Let's keep this open and see what folks say.  @brunoborges Twitter poll?

As a side note - this is what a mailing list would be useful for? I'm not sure how we can broadcast this sort of thing.

dsgrieve:
Moving forward with this could potentially create two code streams where any fix in one would have to be applied to the other.

PR #171, opened 2021-11-08
Document GCToolkit workflow

kirk-microsoft:
Initial skeleton for a developers guide. Initially explains the flow through the toolkit.


dsgrieve (changes requested):
While I have nothing against the API changes, they have nothing to do with the issue at hand. The changes in the .java files should be rolled back and issued in a separate PR.

DEVELOPERS_GUIDE.md
```diff
@@ -0,0 +1,87 @@
...
+
+###Parsers
+
+The parsers extract information from the log files and generation JVMEvents. All JVMEvents share the same 3 characteristics; the type of the event (reflected in the class), the time of the event, and the duration of the event. These properties are immutable and as such, the creation of a JVMEvent is delayed until they are known. The parse rules make heavy use of regular expressions. This is done so that the parser can make a best effort to extract information from a log even if that log is mixed with output from an unexpected source of the log or, the logs lines are some how corrupted.
+
+The currently existing parsers are generational, G1, Z, and Shenandoah. The generational parser cover the serial, parallel (PSYoung, PSOld), ParNew, CMS, and iCMS collectors. There are separate parsers for unified and pre-unified logs produced by the generational and G1 collectors. The pre-unified generational collectors should cope with logs produced by JDK 1.4.2 whereas the G1 log parsers are for JDK 8+. ZGC support starts at JDK 16 whereas Shenandoah starts at JDK 11.
```
dsgrieve, line 35: Just a nit, but 'whereas' is used to compare and contrast - "The banker's hands were smooth and soft whereas the farmer's hands were hard and calloused". Since 'whereas' is a conjunction, you can fix this (if you care) by just removing and making separate sentences.

DEVELOPERS_GUIDE.md
```diff
@@ -0,0 +1,87 @@
...
+
+The flow through the parser starts with it accepting a log line. The line is run through a set of filters designed to recognize if we are interested in parsing it or not. If we are interested, the line is run over a set of regular expressions encapsulated in a class called GCParseRule. If the line hits a GCParseRule, a corresponding parse method is called. The values extracted from the log lines will be stored in a forward reference. This forward reference will be called upon to create the event once all of the lines for that GC event have been scanned. The event is then published on the JVMEvent message bus.
```
dsgrieve, line 37: You might want to emphasize that there is a one-to-one correspondence between a GCParseRule and the parse method.

DEVELOPERS_GUIDE.md
```diff
@@ -0,0 +1,87 @@
...
+
+An example of a GCParseRule can be seen in the code fragment below.
+
+
+GCParseRule CONCURRENT_CYCLE_END = new GCParseRule("CONCURRENT_CYCLE_END", "Concurrent Cycle " + CONCURRENT_TIME);
```
dsgrieve, line 42: Might be helpful to surround the code and log lines with a code block. Also, you might want to state that CONCURRENT_TIME is a regex that matches a real number followed by 'secs' or 'ms', for example 89.437ms. 
Update: And now I see where you define it below. Might be helpful to define it here. Plus, I think the explanation you give is a little too detailed for this. You can always put a link to the code.

DEVELOPERS_GUIDE.md
```diff
@@ -0,0 +1,87 @@
...
+String TIME = "(-?" + REAL_NUMBER + “)";
+String REAL_NUMBER = INTEGER + DECIMAL_POINT + INTEGER;
+String DECIMAL_POINT = "(?:\\.|,)";
+String INTEGER = "\\d+";
+
+Since many lines contain either a pause or concurrent time, defining these “tokens” eases the complexity of creating and debugging the parse rules.
+
+####Testing
+
+There are two main types of tests that are used to QA parsers and rules. The first is known as a rules capture test. In this test the rule is exposed to everything that it is likely to encounter. The test passes if the rule captures everything it should capture and doesn’t capture anything else. In the former test, the parser is configured to send generated events to the test. The test the counts each of the different types of events. The test passes if the counts for each event type reach the expected value.
```
dsgrieve, line 61: "In the former test" needs to be replaced with whatever it is that you call the second type of test.

DEVELOPERS_GUIDE.md
```diff
@@ -0,0 +1,87 @@
...
+
+There is a third test that has, unfortunately, not bee applied to all parsers as of yet. In this test, the parsers will parse a log fragment to produce a well known event. The test passes if all of the attributes of that event match the expected value.
+
+
+For historical reasons, SurvivorSpace is treated as a JVMEvent even though in reality, it’s an attribute of a young generational collection. This record is created by a parser that is specialized to parse the log and extract out the survivor information.
```
dsgrieve, line 66: Why is this here? Shouldn't this be under JVMEvent section?

DEVELOPERS_GUIDE.md
```diff
@@ -0,0 +1,87 @@
...
+String DECIMAL_POINT = "(?:\\.|,)";
+String INTEGER = "\\d+";
+
+Since many lines contain either a pause or concurrent time, defining these “tokens” eases the complexity of creating and debugging the parse rules.
+
+####Testing
+
+There are two main types of tests that are used to QA parsers and rules. The first is known as a rules capture test. In this test the rule is exposed to everything that it is likely to encounter. The test passes if the rule captures everything it should capture and doesn’t capture anything else. In the former test, the parser is configured to send generated events to the test. The test the counts each of the different types of events. The test passes if the counts for each event type reach the expected value.
+
+There is a third test that has, unfortunately, not bee applied to all parsers as of yet. In this test, the parsers will parse a log fragment to produce a well known event. The test passes if all of the attributes of that event match the expected value.
```
dsgrieve, line 63: s/third test/third type of test/

DEVELOPERS_GUIDE.md
```diff
@@ -0,0 +1,87 @@
...
+
+![GCToolkit Design](images/GCToolkit_Events.png)
+Figure 2. GCToolkit Event Hierarchy
+
+### Aggregator and Aggregation
+
+
+As can be seen in Figure 1, the end point for delivery of a JVMEvent as far as the framework is concerned, is an Aggregator. Aggregator classes must be provided by the users of GCToolkit as they are not provided as they are considered out of scope for this project. The role of the Aggregator is to capture the JVMEvent objects that are of interest to it and the extract the data of interest and pass it along to an Aggregation.
```
dsgrieve, line 78: s/JVMEvent as far/JVMEvent, as far/ or reword as 'As far as the framework is concerned, the endpoint for delivery of a JVMEvent is an Aggregator, as can be seen in Figure 1.'
I would move the last sentence before the sentence about Aggregator classes being out of scope and the sentence in the next paragraph about Aggregations being supplied by the user into a separate paragraph at the end of this section. And then point the reader to the sample.

DEVELOPERS_GUIDE.md
```diff
@@ -0,0 +1,87 @@
...
+
+As is the case with Aggregator, Aggregation classes must be supplied by users of GCToolkit. The role of an Aggregation is to collect the data and perform all calculations needed to transform the event data into a form that fits the needs of the client. For example, a GUI may want to display heap occupancy after collection. The Aggregator would capture all GC pause events from which contain reasonable accurate measures of this metrics. This information would be collected into an Aggregator.
```
dsgrieve, line 80: s/events from which contain/events, which contain/ 
s/metrics/metric/
s/Aggregator/Aggregation/ in the last sentence

DEVELOPERS_GUIDE.md
```diff
@@ -0,0 +1,87 @@
...
+
+    It is the responsibility of GCToolkit to load and register all instances of Aggregation via the module’s SPI
```
dsgrieve, line 82: Need to mention the Aggregates and Collates annotations.

I would say that Aggregation classes need to be registered with GCToolkit, and that this can be done via GCToolkit API or through SPI. The benefit of SPI is that it allows the user to add additional Aggregations without having to modify code.

dsgrieve (changes requested):

api/src/main/java/com/microsoft/gctoolkit/GCToolKit.java
```diff
@@ -23,22 +26,24 @@
 
     private static final Logger LOGGER = Logger.getLogger(GCToolKit.class.getName());
 
-    private static JavaVirtualMachine loadJavaVirtualMachine() {
+    private JavaVirtualMachine loadJavaVirtualMachine(GCLogFile logFile) {
+        //todo: change to using ServiceLoader...
+        final String className = (logFile.isUnified()) ? "com.microsoft.gctoolkit.vertx.jvm.UnifiedJavaVirtualMachine" : "com.microsoft.gctoolkit.vertx.jvm.PreUnifiedJavaVirtualMachine";
```
johnoliver, line 31: Structural issue here is probably that constructing JavaVirtualMachines should not be performed in the `api` this code means that there is a runtime dependency between the `api` and `vertx` but not enforced by maven, Fix probably should be to move JavaVirtualMachine construction to a module that has visibility of both `vertx` and `api`

dsgrieve: The reason it is like this was just an expedient to getting gctoolkit out in the open. Too much happens in that DefaultJavaVirtualMachine. JavaVirtualMachine impl should be decoupled from vertx. If the vertx layer needs JavaVirtualMachine, it should be injected. Or the API needed from vertx should be abstracted and injected into GCToollKit.

dsgrieve:
@kcpeppe It looks like you've mixed the refactoring from PR #173 into this one.

kirk-microsoft:
@dsgrieve, it's a continuation.

DEVELOPERS_GUIDE.md
```diff
@@ -0,0 +1,87 @@
...
+
+A second feature of Vert.x that GCToolkit is dependent on is the construct of a Verticle. A verticle acts as a container to encapsulate a component designed to process events. In GCToolkit vertices are used to encapsulate the DataSource, parsers, and aggregators. Vert.x takes care of many of the concerns that one has when distributing the processing events. In addition, verticals are a covenant abstraction that nicely supports strong separation of concerns.
+
+ 
+### Data Source 
+
+DataSource is an interface that allows one to define specializations for a source of GC events that maybe of interest. Currently the toolkit contains two data sources, a FileDataSource and a SafepointogFile. A GClogFile is a specialization of a FileDataSource. This is in turn has two specializations, SingleGCLogFIle and RotatingGCLogFile. The concrete classes take care of the details of being able to create a continuous stream of their contents.
```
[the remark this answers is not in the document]
kirk-microsoft: done

DEVELOPERS_GUIDE.md
```diff
@@ -0,0 +1,87 @@
...
+
+Another role of the DataSource is to provide a Diary (or summary) of important features found in the data. This would include information such as, is this a unified log or a pre-unified log, which version of the JVM produced the file. What flags where used to produce the file, which collector (combinations) are in play and so on. This information is used to construct the data paths through GCToolkit. It may also be used by clients of GCToolkit for their own purposes. For example, analytics may use the information to modulate how they process data.
+
+An example of this would be an analytic that looked at System.gc() behaviour. In JDK 1.7.0_40 the logs stopped reporting user triggered collections unless a new flag was enabled. This left the analytic blind to these events. In JDK 1.8.0, this new flag was automatically turned on when the PrintGCDetails flag was set. Making the analytics aware of the JDK version allowed them to calculate when System.gc() was likely being called thus enabling the analysis when the cause was missing in the log.
```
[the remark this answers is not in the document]
kirk-microsoft: fixed

dsgrieve:
This PR should be abandoned, right?

kirk-microsoft:
I'll diff it to make sure I've captured every thing afterwards, yes, I'll put it to rest

karianna:
@kirk-microsoft do you have a separate docs PR or want me to create one with these contents?

Issue #183, opened 2021-12-15
@Aggregates({EventSource.SAFEPOINT}) will fall into dead loop


ZhangShushu123:
When I create an Aggregator with @Aggregates({EventSource.SAFEPOINT}). The program will loop forever.
I read the source code and I found that there is no LogFileParser with the outbox _SAFEPOINT_ . 
Therefore, when the program execute the code `aggregatorVerticles.forEach(AggregatorVerticle::awaitCompletion);` , the Aggregator with @Aggregates({EventSource.SAFEPOINT}) will not complete forever.

ZhangShushu123:
Is GCToolkitVertxParametersForPreUnifiedLogs.java line 140 `AggregatorVerticle aggregatorVerticle = new AggregatorVerticle(SAFEPOINT_OUTBOX);` change to `AggregatorVerticle aggregatorVerticle = new AggregatorVerticle(GCToolkitVertx.JVM_EVENT_PARSER_OUTBOX);`? 
Because there is no AggregatorVerticle with the inbox _JVM_EVENT_. 
However, I think we should add an `EventSource` enum value of _JVM_EVENT_ will be better than the before one.

dsgrieve:
Nice find! 

I think the issue here is that the SafePointParser is never created. I think part of the fix is to add this block of code to the `initLogFileParsers` method of GCToolkitVertxParametersForPreUnifiedLogs. 
```java
        if (diary.isApplicationStoppedTime() || diary.isApplicationRunningTime()) {
            logFileParsers.add(new LogFileParser(GCToolkitVertx.PARSER_INBOX, SAFEPOINT_OUTBOX, consumer -> new SafepointParser(diary, consumer)));
        }
```
The `if (diary.isApplicationStoppedTime() || diary.isApplicationRunningTime())` test is probably needed around the code that calls `getAggregators(EventSource.SAFEPOINT, registeredAggregations)` down in `initAggregatorVerticles` in the same class.

It would be awesome, @ZhangShushu123, if you could work this issue!

ZhangShushu123:
By the way, I think adding code to remove `aggregatorVerticles` which inbox can not match with the `logFileParser `in `aggregateDataSource()` is a better way, because it will prevent the dead loop. Moreover, logging the removed `aggregatorVerticle`  in console with error level will tell us which `aggregatorVerticle` can not match and we should fix in the future.

ZhangShushu123:
[...]

PR #189, opened 2021-12-18
fix dead loop(#183)

ZhangShushu123:
fix dead loop(#183)


ZhangShushu123:
When will it be merged into main branch?

karianna:
@ZhangShushu123 Thank you so much for this PR! Apologies for the delay.  Most of the committers are off on holiday leave until the first week of January.

kirk-microsoft:
Hi @ ZhangShushu123, Thank you for this patch and my apologies for the delay in addressing this. I did look at this as soon as it came in. I was battling getting 2.0.5 out and maven central wasn't cooperating and given the bits you were touching, I wanted to engage in a deeper review which I simply didn't have the cycles for. Now I have...

As an FYI, the issue (not with your patch) with this chunk of code is that the parsers have direct knowledge of the message bus which I feel is a violation of SRP (Single Responsibility Principle). In the implementation that we had prior to the modularization refactoring separated this behaviour into a message bus connector object that wrapped the parser. That refactoring is likely going to take some time.

The nice thing about your patch is that it addresses another issue and that is that all of the aggregators are registered irregardless of their applicability to the log being parsed. So, lets kill the dead loop code and deal with the other stuff later.

ZhangShushu123:
Thank you very much for taking the time to merge code. And I look forward to your next refactoring work.

PR #202, opened 2022-01-28
fix: add UnifiedJavaVirtualMachine to modulo-info

leveretconey:
Hi,

Could I have a review of this fix?

UnifiedJavaVirtualMachine should be provided as a service in vertx module, otherwise UnifiedJavaVirtualMachine can not be loaded by GCToolKit::loadJavaVirtualMachine.

Thanks,
Yifeng

### DIFF
--- a/vertx/src/main/java/module-info.java
+++ b/vertx/src/main/java/module-info.java
@@ -2,6 +2,7 @@
 // Licensed under the MIT License.
 
 import com.microsoft.gctoolkit.vertx.jvm.PreUnifiedJavaVirtualMachine;
+import com.microsoft.gctoolkit.vertx.jvm.UnifiedJavaVirtualMachine;
 
 /**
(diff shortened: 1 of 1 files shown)


vertx/src/main/java/module-info.java
```diff
@@ -17,6 +18,7 @@
             com.microsoft.gctoolkit.api;
 
     provides com.microsoft.gctoolkit.jvm.JavaVirtualMachine with
-            PreUnifiedJavaVirtualMachine;
+            PreUnifiedJavaVirtualMachine,
```
dsgrieve, line 21: @leveretconey - Would you mind getting rid of the import statements and using the fully qualified class names in the provides statement? Using fully qualified class names rather than imports is the accepted style for module-info files.

dsgrieve (approved):

PR #226, opened 2022-09-06
[zgc][parser] Ensure that ZGCMemoryPoolSummary and OccupancySummary use kilobytes

j-bahr:
Ensure that ZGC memory values are normalized to kilobytes. The code in its current form will return a double without taking into account the units associated with the memory value. This can lead to multiple GC cycles that report odd results. This diff ensures that as they are parsed the are normalized to a long value in terms of kilobytes.  The regex's already capture the units so we'll simply scale the constructor inputs to ZGCMemoryPoolSummary and OccupancySummary by the captured unit string.  We'll use the helper method `.toKBytes(...)` which is available to us in the the trace class already.

### DIFF
--- a/parser/src/main/java/com/microsoft/gctoolkit/parser/ZGCParser.java
+++ b/parser/src/main/java/com/microsoft/gctoolkit/parser/ZGCParser.java
@@ -31,7 +31,6 @@ import java.util.logging.Logger;
  * CMS failures
  * System.gc() calls
  */
-
 public class ZGCParser extends UnifiedGCLogParser implements ZGCPatterns {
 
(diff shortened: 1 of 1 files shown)


j-bahr:
Please let me know if there is a project code formatting standard, I'd be happy to reformat to meet the project standards. While working the editor cleaned up a number of spacing issues which lead to a slight larger PR than expected. I wish this could have been picked up in a separate, format only dirr. Anyways. Please let me know if I can better address this.

kcpeppe (approved):

kcpeppe:
Admiditly support for ZGC isn't well tested, we need more unit testing in this area. So this looks very good. Thank you! @j-bahr.

PR #228, opened 2022-09-07
[ZGC] Improve parsing for Java 17 style output

j-bahr:
ZGC was marked production ready in Java 17. There were a number of changes made to the ZGC log output in the past 2-3 years that have made it into the 17 release. I've updated the tests with samples from a Java 17 run on a sample app to test changes made to the ZGCCycle class.  Java sample app was invoked with the following flag `-Xlog:gc=info,gc+start=info,gc+phases=info,gc+load=info,gc+mmu=info,gc+marking=info,gc+metaspace=info,gc+nmethod=info,gc+ref=info,gc+reloc=info,gc+heap=info:file=/logs/gclog:level,tags,uptime:filecount=10,filesize=250m`

* elastic metaspace was introduced in this commit https://github.com/openjdk/zgc/commit/7ba6a6bf003b810e9f48cb755abe39b1376ad3fe#diff-9aa1419e172b665d42f187e7f9d46753f396ab6e0d192906d659ff81b3b6cf12 which removes the capacity field. A specific metaspace class was added to reflect these changes.

* Memory pool summary no longer emits reserved heap, removed to reflect changes in this file https://github.com/openjdk/zgc/blob/1f8813495e8184b6c38319df1c2cb70de7811a76/src/hotspot/share/gc/z/zStat.cpp#L1432-L1503

* MarkFree phase was added and exported in the ZGCCycle summary. May be opportunity to add Mark Continue phase in the future. Phase are defined here https://github.com/openjdk/zgc/blob/26c03c1860c6da450b5cd6a46576c78bea682f96/src/hotspot/share/gc/z/zDriver.cpp#L45-L54

* Reclaimed and MemorySummary were updated to export value in kilobytes based on the unit strings

* Relocation summary is not yet implemented so it was removed form the ZGCParserRulesTest class as the old capture doesn't work with the new format. I'll update this in a separate PR.

* ZGCParserTest was update with a new format, along with all test values, that were hand calculate to ensure we're accurately parsing the data.

### DIFF
--- a/api/src/main/java/com/microsoft/gctoolkit/event/zgc/ZGCCycle.java
+++ b/api/src/main/java/com/microsoft/gctoolkit/event/zgc/ZGCCycle.java
@@ -30,6 +30,9 @@ public class ZGCCycle extends GCEvent {
     private DateTimeStamp concurrentMarkTimeStamp;
     private double concurrentMarkDuration;
 
+    private DateTimeStamp concurrentMarkFreeTimeStamp;
+    private double concurrentMarkFreeDuration;
+
     private DateTimeStamp pauseMarkEndTimeStamp;
     private double pauseMarkEndDuration;
 
@@ -47,7 +50,6 @@ public class ZGCCycle extends GCEvent {
 
     private DateTimeStamp concurrentRelocateTimeStamp;
     private double concurrentRelocateDuration;
-    private ZGCMemoryPoolSummary metaspace;
 
     private double[] load = new double[3];
     private double[] mmu = new double[6];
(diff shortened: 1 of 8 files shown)


j-bahr:
One of the motivations here for upgrading for the latest output format is because ZGC was not really considered "production ready" in an LTS release until Java 17 (from Per Liden's ZGC talk) . I suspect that the majority of users who want leverage ZGC would be doing it on Java 17+ not Java 11 given its less than ideal performance on Java 11. Please let me know if you disagree with this assertion.

karianna:
@kirk-microsoft - Thoughts? I'm pretty sympathetic to this line of thinking (ZGC for 17+ only).

karianna (approved):
Code changes LGTM

PR #229, opened 2022-09-08
[ZGC] Add GC ID to the ZGCCycle

j-bahr:
The GC id can be useful when analyzing data and may be valuable to users. It can help order data later or give an indication of how many GC cycles have happened since since the process start or between points in time.

I've plumbed the GC id from the ZGCParser through to the ZGCCycle class. It'll be captured as part of the cycle start regex.

### DIFF
--- a/api/src/main/java/com/microsoft/gctoolkit/event/zgc/ZGCCycle.java
+++ b/api/src/main/java/com/microsoft/gctoolkit/event/zgc/ZGCCycle.java
@@ -24,6 +24,8 @@ public class ZGCCycle extends GCEvent {
         super(timeStamp, gcType, duration);
     }
 
+    private long gcId;
+
     private DateTimeStamp pauseMarkStartTimeStamp;
     private double pauseMarkStartDuration;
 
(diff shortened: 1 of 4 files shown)


karianna:
@j-bahr - one failing test? ZGCParserTest.infoLevelZGCCycle:126 expected: <true> but was: <false>

karianna:
@kirk-microsoft over to you for a review also

kirk-microsoft:
The failing test is on me, it's not been fixed in my pull request. GCID is a new(ish) field and as such, it's not been included in any of the GC events. Adding it to ZGC makes that event asymmetric but if it's going to be added, we need to start some where.... LGTM

j-bahr:
@kirk-microsoft, I think the failure was on my end, I merged main into my branch and must have run the tests on a previous build. It's a simple fix and I can push an update shortly.

I've been heavily focused on ZGC as it's the collector I've been profiling this week. I may plan to profile other GC's like shenandoah and G1 in which case I could put up a PR adding the GC Id to those to make them more symmetric. Hopefully I get time to add it to the other collectors.

j-bahr:
Oops and I just noticed that this was already merged with fixes. Thanks!

kirk-microsoft:
It was a simple change to the GCID that the test was looking for so no worries.
@j-bahr : do you have "interesting" zgc logs that you could contribute to our GCToolKit-Testdata project?

j-bahr:
@kirk-microsoft, possibly, I'll likely need to get approval first before contributing non-code, log samples. I'll have to get back to you on that.

PR #230, opened 2022-09-08
Consolidate utility method

kirk-microsoft:
Consolidate utility method to toKBytes() along with a number refactorings.

### DIFF
--- a/api/src/main/java/com/microsoft/gctoolkit/message/JVMEventConsumer.java
+++ b/api/src/main/java/com/microsoft/gctoolkit/message/JVMEventConsumer.java
@@ -1,0 +1,4 @@
+package com.microsoft.gctoolkit.message;
+
+public interface JVMEventConsumer {
+}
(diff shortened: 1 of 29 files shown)


api/src/main/java/com/microsoft/gctoolkit/message/JVMEventConsumer.java
```diff
@@ -0,0 +1,4 @@
+package com.microsoft.gctoolkit.message;
+
+public interface JVMEventConsumer {
```
dsgrieve, line 3: From the name of the interface, I expected something like `extends java.util.function.Consumer<JVMEvent>`

parser/src/main/java/com/microsoft/gctoolkit/parser/GCLogParser.java
```diff
@@ -175,24 +175,24 @@ CPUSummary extractCPUSummary(String line) {
...
-            case 'B':
-            case 'b':
-                value /= 1024;
-                break;
-            default:
-        }
-
-        return value;
-    }
+//    double toKBytes(double value, String units) {
```
dsgrieve, line 178: Remove

PR #248, opened 2022-12-20
An improvement on ParserTest to fix diarization

dsgrieve:
@kcpeppe - I took a swipe at fixing the diarization issue in ParserTest. Basically, the setUp() method didn't have access to the String[] lines to create the diary. I pushed setUp() up into ParserTest and introduced two abstract methods - one to get the Diarizer, the other to get the GCLogParser. I then changed the feedParser method so that it only takes the String[] lines. All of the diarization and channel setup is done in feedParser(String lines), which also cleans up the redundant code in the three tests that use ParserTest at the moment.


karianna (approved):
LGTM on the micro level

kirk-microsoft (commented):
Looks great!

PR #249, opened 2022-12-30
Tidied up the map and iteration

kabutz:


karianna (approved):
LGTM - @kirk-microsoft if collectionCycleCounts was effectively threadsafe then we should make that clear else I think this is a sane change that adds safety.

kirk-microsoft:
TLDR; if you don't know you should travel the safe route and this is the safe route. Processing a events is single threaded.

Some details: Aggregator and as a consequence, Aggregation exist in a Vert.x Verticle ensuring that the processing events was single threaded. However, if a reader thread accessed the Aggregation, that would be an unsafe read. Having said that, it is unclear if that matters as in this case the reader would be reading values at some random point in time w.r.t. the event publication timeline. Having said that, reading from a HashMap that is being written to creates a very small window of opportunity where the reader thread could be trapped in an infinite loop. However, we have not seen this happening since the collision resolution mechanism changed from LinkedList to AVL-Tree (JDK 9?).

kabutz:
> TLDR; if you don't know you should travel the safe route and this is the safe route. Processing a events is single threaded.
> 
> Some details: Aggregator and as a consequence, Aggregation exist in a Vert.x Verticle ensuring that the processing events was single threaded. However, if a reader thread accessed the Aggregation, that would be an unsafe read. Having said that, it is unclear if that matters as in this case the reader would be reading values at some random point in time w.r.t. the event publication timeline. Having said that, reading from a HashMap that is being written to creates a very small window of opportunity where the reader thread could be trapped in an infinite loop. However, we have not seen this happening since the collision resolution mechanism changed from LinkedList to AVL-Tree (JDK 9?).

The red-black tree is used since Java 8, and yes, infinite loops do still happen. I also fixed the iteration, which was over a keySet with a lookup, rather than an entrySet.

PR #254, opened 2023-01-15
Got rid of Phaser and replaced with CountDownLatch

kabutz:


kirk-microsoft:
Was trying to avoid having to directly count registered aggregators but.... lets see how this goes.

PR #269, opened 2023-02-03
Phaser

kirk-microsoft:
API added to allow clients to use GCToolKit on the classpath

### DIFF
--- a/IT/src/test/java/com/microsoft/gctoolkit/integration/NoModuleIntegrationTest.java
+++ b/IT/src/test/java/com/microsoft/gctoolkit/integration/NoModuleIntegrationTest.java
@@ -22,11 +22,9 @@ import java.util.Optional;
 
 import static org.junit.jupiter.api.Assertions.fail;
 
+@Tag("classPath")
 public class NoModuleIntegrationTest {
 
-    //Todo: turn on this test once mvn has been configured to run it with all of the supporting bits
-    // on the classpath and not the module-path
-    @Tag("classPath")
     @Test
     public void testMain() {
         Path path = new TestLogFile("cms/defnew/details/defnew.log").getFile().toPath();
--- a/IT/src/test/java/com/microsoft/gctoolkit/integration/NoModuleIntegrationTest.java
+++ b/IT/src/test/java/com/microsoft/gctoolkit/integration/NoModuleIntegrationTest.java
@@ -14,6 +14,7 @@ import com.microsoft.gctoolkit.vertx.VertxDataSourceChannel;
 import com.microsoft.gctoolkit.vertx.VertxJVMEventChannel;
 import org.junit.jupiter.api.Assertions;
 import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.Tag;
 
 import java.io.IOException;
 import java.nio.file.Path;
@@ -23,6 +24,9 @@ import static org.junit.jupiter.api.Assertions.fail;
 
 public class NoModuleIntegrationTest {
 
+    //Todo: turn on this test once mvn has been configured to run it with all of the supporting bits
+    // on the classpath and not the module-path
+    //@Tag("classPath")
     //@Test
     public void testMain() {
         Path path = new TestLogFile("cms/defnew/details/defnew.log").getFile().toPath();
--- a/IT/src/test/java/com/microsoft/gctoolkit/integration/NoModuleIntegrationTest.java
+++ b/IT/src/test/java/com/microsoft/gctoolkit/integration/NoModuleIntegrationTest.java
@@ -1,0 +1,116 @@
+package com.microsoft.gctoolkit.integration;
+
+import com.microsoft.gctoolkit.GCToolKit;
+import com.microsoft.gctoolkit.integration.aggregation.CollectionCycleCountsSummary;
+import com.microsoft.gctoolkit.integration.aggregation.HeapOccupancyAfterCollectionSummary;
+import com.microsoft.gctoolkit.integration.aggregation.PauseTimeSummary;
+import com.microsoft.gctoolkit.integration.io.TestLogFile;
+import com.microsoft.gctoolkit.io.GCLogFile;
+import com.microsoft.gctoolkit.io.SingleGCLogFile;
+import com.microsoft.gctoolkit.jvm.JavaVirtualMachine;
+import com.microsoft.gctoolkit.parser.CMSTenuredPoolParser;
+import com.microsoft.gctoolkit.parser.GenerationalHeapParser;
+import com.microsoft.gctoolkit.vertx.VertxDataSourceChannel;
+import com.microsoft.gctoolkit.vertx.VertxJVMEventChannel;
+import org.junit.jupiter.api.Assertions;
+import org.junit.jupiter.api.Test;
+
+import java.io.IOException;
+import java.nio.file.Path;
+import java.util.Optional;
+
+import static org.junit.jupiter.api.Assertions.fail;
+
+public class NoModuleIntegrationTest {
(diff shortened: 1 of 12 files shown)


IT/src/test/java/com/microsoft/gctoolkit/integration/NoModuleIntegrationTest.java
```diff
@@ -0,0 +1,120 @@
...
+import java.nio.file.Path;
+import java.util.Optional;
+
+import static org.junit.jupiter.api.Assertions.fail;
+
+public class NoModuleIntegrationTest {
+
+    //Todo: turn on this test once mvn has been configured to run it with all of the supporting bits
+    // on the classpath and not the module-path
+    @Tag("classPath")
```
dsgrieve, line 29: This @Tag could/should be at the class level.

api/src/main/java/com/microsoft/gctoolkit/GCToolKit.java
```diff
@@ -127,13 +88,136 @@ public void loadAggregationsFromServiceLoader() {
      * @see Aggregation
      * @see JavaVirtualMachine
      */
-    public void registerAggregation(Aggregation aggregation) {
+    public void loadAggregation(Aggregation aggregation) {
```
dsgrieve, line 91: Yikes. This is a breaking change on top of the breaking change from version 2.x.y (where the arg was a Class<? extends Aggregation>).

api/src/main/java/com/microsoft/gctoolkit/GCToolKit.java
```diff
@@ -127,13 +88,136 @@ public void loadAggregationsFromServiceLoader() {
...
+                        .stream()
+                        .map(ServiceLoader.Provider::get)
+                        .findFirst()
+                        .orElseThrow(() -> new ServiceConfigurationError("Internal Error - No suitable DataSourceBus implementation found")));
+            } else {
+                try {
+                    Class clazz = forName("com.microsoft.gctoolkit.vertx.VertxDataSourceChannel", true, Thread.currentThread().getContextClassLoader());
+                    loadDataSourceChannel((DataSourceChannel) clazz.getConstructors()[0].newInstance());
+                } catch (Exception e) {
+                    throw new ServiceConfigurationError("Unable to find a suitable provider to create a diary");
```
dsgrieve, line 124: Wrong message?

api/src/main/java/com/microsoft/gctoolkit/GCToolKit.java
```diff
@@ -127,13 +88,136 @@ public void loadAggregationsFromServiceLoader() {
...
+                }
+            }
+        }
+    }
+
+    public void loadJVMEventChannel(JVMEventChannel channel) {
```
dsgrieve, line 130: Why "load" and not "set"?

api/src/main/java/com/microsoft/gctoolkit/GCToolKit.java
```diff
@@ -127,13 +88,136 @@ public void loadAggregationsFromServiceLoader() {
...
+                    "com.microsoft.gctoolkit.parser.UnifiedSurvivorMemoryPoolParser",
+                    "com.microsoft.gctoolkit.parser.ZGCParser"
+            };
+            try {
+                dataSourceParsers = Arrays.stream(parsers)
+                        .map(parserName -> {
+                            try {
+                                return forName(parserName, true, Thread.currentThread().getContextClassLoader());
+                            } catch (ClassNotFoundException e) {
+                                throw new ServiceConfigurationError("Unable to find a suitable provider to create a DataSourceParser");
```
dsgrieve, line 195: I suggest replacing the whole try-catch block starting on line 189 with this: 
```java
dataSourceParsers = Arrays.stream(parsers)
                        .map(parserName -> {
                            try {
                                Class<?> clazz = forName(parserName, true, Thread.currentThread().getContextClassLoader());
                                return Optional.of(clazz.getConstructors()[0].newInstance());
                            } catch (ClassNotFoundException 
                                     | InstantiationException 
                                     | IllegalAccessException 
                                     | InvocationTargetException e) {
                                return Optional.empty();
                            }
                        })
                        .filter(Optional::isPresent)
                        .map(optional -> (DataSourceParser) optional.get())
                        .filter(consumer -> consumer.accepts(diary))
                        .collect(Collectors.toList());
                if (dataSourceParsers.isEmpty()) {
                    throw new ServiceConfigurationError("Unable to find a suitable provider to create a DataSourceParser");
                } 
```
which would loop through the whole parsers array instead of bailing on the first exception.

IT/src/test/java/com/microsoft/gctoolkit/integration/ZeroAggregationTest.java
```diff
@@ -47,18 +47,14 @@ public void testNoAggregationRegistered() {
...
-        /*
-         * GC log files can come in  one of two types: single or series of rolling logs.
-         * In this sample, we load a single log file.
-         * The log files can be either in text, zip, or gzip format.
-         */
         GCLogFile logFile = new SingleGCLogFile(path);
         GCToolKit gcToolKit = new GCToolKit();
-        // Load our local Aggregation that will not registered for the given log file
-        gcToolKit.registerAggregation(new ZeroAggregationTest.TestAggregation());
+        /* Load our local Aggregation that will not registered for the given log file */
```
karianna, line 56:
```suggestion
        // Load our local Aggregation that will not be registered for the given log file
```

api/src/main/java/com/microsoft/gctoolkit/GCToolKit.java
```diff
@@ -127,13 +88,136 @@ public void loadAggregationsFromServiceLoader() {
...
+                    .collect(Collectors.toList());
+        } else {
+            dataSourceParsers = new ArrayList<>();
+            dataSourceParsers.addAll(registeredDataSourceParsers);
+        }
+
+
+        if (dataSourceParsers.isEmpty()) {
+            String[] parsers = {
+                    "com.microsoft.gctoolkit.parser.CMSTenuredPoolParser",
```
karianna, line 177: Should we externalize this in a properties file?

api/src/main/java/com/microsoft/gctoolkit/GCToolKit.java
```diff
@@ -127,13 +88,136 @@ public void loadAggregationsFromServiceLoader() {
...
+                            try {
+                                return (DataSourceParser) clazz.getConstructors()[0].newInstance();
+                            } catch (Exception e) {
+                                throw new ServiceConfigurationError("Unable to find a suitable provider to create a DataSourceParser");
+                            }
+                        })
+                        .filter(consumer -> consumer.accepts(diary))
+                        .collect(Collectors.toList());
+            } catch (Exception e) {
+                e.printStackTrace();
```
karianna, line 208: use logger

api/src/main/java/com/microsoft/gctoolkit/GCToolKit.java
```diff
@@ -127,13 +88,136 @@ public void loadAggregationsFromServiceLoader() {
...
+            }
+        }
+        for (DataSourceParser dataSourceParser : dataSourceParsers) {
```
karianna, line 211: SRP?  Maybe this belongs in its own method

karianna (changes requested):

api/src/main/java/com/microsoft/gctoolkit/GCToolKit.java
kirk-microsoft: Not a breaking change, just an addition.

PR #270, opened 2023-02-03
split class-path and module-path test goals into separate poms

dsgrieve:


dsgrieve:
Commit cd889cd0c4e484f763d3ecf152615ae5985eac7b seems to work, so this change isn't needed.

PR #288, opened 2023-04-07
GCToolKit#addDataSourceParser

dsgrieve:
Using loadDataSourceParser causes the code to skip loading DataSourceParsers using the SPI. We added addDataSourceParser which adds a DataSourceParser to the ones that are loaded with the SPI. Along the way, we noticed that the Aggregators are loaded in such a way that they might never actually be used, which would lead to a hang because the Phaser in AbstractVirtualMachine might never be decremented for those Aggregators that are never used. So, a good deal of this change is to fix that issue, which we did by having the DataSourceParser report what events it produces (added DataSourceParser#producesEvents()). Now the code can check if any DataSourceParsers produce the events that an Aggregator aggregates. If not, the Aggregator is not used. 

So... The files to focus on are GCToolkit and AbstractJavaVirtualMachine


api/src/main/java/com/microsoft/gctoolkit/aggregator/Aggregator.java
```diff
@@ -121,10 +121,8 @@ public void onCompletion(Runnable task) {
      * Call a callback when aggregation is completed.
      */
     private void complete() {
-        Runnable t = completionTask;
-        this.completionTask = null;
-        if (t != null)
-            Executors.newSingleThreadExecutor().execute(t);
+        if (completionTask != null)
```
dsgrieve, line 124: The reason for this change is that it is possible for an Aggregator to get events from more than one EventSource. If the completionTask is nulled out, then the Phaser only gets decremented by the first EventSource but not the subsequent EventSources. 

The test CaptureJVMTerminationEventTest was hanging because of this issue.

api/src/main/java/com/microsoft/gctoolkit/GCToolKit.java
```diff
@@ -209,11 +246,19 @@ private void loadDataSourceParsers(Diary diary) {
             }
         }
 
+        /*
+         *  add in any additional parsers not provided by the module SPI.
```
karianna, line 250: Could be a single line comment

parser/src/test/java/com/microsoft/gctoolkit/parser/ConcurrentMarkSweepPhaseParserRulesTest.java
```diff
@@ -30,13 +32,21 @@ public void testCMSParseRules() {
 
     /* Code that is useful when testing individual records */
 
-    private final boolean debugging = Boolean.getBoolean("microsoft.debug");
+    private static final boolean DEBUGGING;
```
karianna, line 35: We seem to be duplicating this code block in a lot of places..

karianna (changes requested):
I think DRY is being broken a little with the debugging implementation

parser/src/test/java/com/microsoft/gctoolkit/parser/ConcurrentMarkSweepPhaseParserRulesTest.java
dsgrieve: True. I replaced all that with a GCToolkit.LOG_DEBUG_MESSAGE(Supplier<String>) method. Using supplier means there is no eval of the code that builds the message unless debugging is enabled.

api/src/main/java/com/microsoft/gctoolkit/GCToolKit.java
```diff
@@ -35,6 +42,38 @@ public class GCToolKit {
...
+     * or contains "className" but does not contain "-className".
+     * For example, to enable debug logging for all classes except UnifiedG1GCParser:
+     * <code>-Dgctoolkit.debug=all,-com.microsoft.gctoolkit.parser.UnifiedG1GCParser</code>
+     *
+     * @param message Supplies the message to log. If null, nothing will be logged.
+     */
+    public static void LOG_DEBUG_MESSAGE(Supplier<String> message) {
+        if (DEBUGGING && message != null) {
+            StackTraceElement[] stackTrace = Thread.currentThread().getStackTrace();
+            String methodName = stackTrace[2].getMethodName();
```
johnoliver, line 67: Is there something special about the 3rd element of a stack trace?

api/src/main/java/com/microsoft/gctoolkit/message/DataSourceParser.java
```diff
@@ -1,9 +1,13 @@
...
+import com.microsoft.gctoolkit.aggregator.EventSource;
 import com.microsoft.gctoolkit.jvm.Diary;
 
+import java.util.Set;
+
 public interface DataSourceParser extends DataSourceChannelListener {
     void publishTo(JVMEventChannel channel);
     void diary(Diary diary);
     boolean accepts(Diary diary);
+    Set<EventSource> producesEvents();
```
johnoliver, line 12: Grammatically, `eventsProduced` might be better

parser/src/test/java/com/microsoft/gctoolkit/parser/ConcurrentMarkSweepPhaseParserRulesTest.java
```diff
@@ -30,26 +33,26 @@ public void testCMSParseRules() {
...
         GCLogTrace trace = rule.parse(string);
         assertNotNull(trace);
-        if (dump) {
-            LOGGER.fine("matches groups " + trace.groupCount());
+        // Enable debugging by setting gctoolkit.debug to true
+        GCToolKit.LOG_DEBUG_MESSAGE(() -> {
+            StringBuilder sb = new StringBuilder("matches groups " + trace.groupCount());
             for (int i = 0; i <= trace.groupCount(); i++) {
-                LOGGER.fine(i + ": " + trace.getGroup(i));
+                sb.append(System.lineSeparator()).append(i).append(": ").append(trace.getGroup(i));
```
johnoliver, line 52: possibly better to use a Joiner

api/src/main/java/com/microsoft/gctoolkit/message/DataSourceParser.java
dsgrieve: 'A DataSourceParser produces events of these types,' versus 'The events produced by a DataSourceParser are of these types'. IDK. I still lean toward producesEvents. 

But then... dataSourceParsers.stream().map(DataSourceParser::producesEvents) versus dataSourceParsers.stream().map(DataSourceParser::eventsProduced) makes me lean more toward eventsProduced. 

@karianna @kirk-microsoft Please weigh in.

api/src/main/java/com/microsoft/gctoolkit/GCToolKit.java
```diff
@@ -35,6 +42,38 @@ public class GCToolKit {
...
+            String className = stackTrace[2].getClassName();
+            String fileName = stackTrace[2].getFileName();
+            int lineNumber = stackTrace[2].getLineNumber();
+            if (isDebugging(className)) {
+                System.out.println(String.format("DEBUG: %s.%s(%s:%d): %s", className, methodName, fileName, lineNumber, message.get()));
```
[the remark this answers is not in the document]
dsgrieve: Good question. My original cut at this was to pass a Logger to this method and I could go back to that very easily. This does raise other questions, such as at what level to log -- there is no Level.DEBUG -- and what to do if someone passes a null Logger -- use anonymous logger? The biggest problem, for me, was that the debug lines got lost in the logger noise. 

I'm not wedded to this method at all. In general, prints to stdout/stderr that depend on a debug flag should be turned into Logger messages.

parser/src/test/java/com/microsoft/gctoolkit/parser/ConcurrentMarkSweepPhaseParserRulesTest.java
dsgrieve: Yeah, that didn't work out so well because of the loop index. Went with this instead:
`sb.append(String.format("%n%d : %s", i, trace.getGroup(i))) `

api/src/main/java/com/microsoft/gctoolkit/GCToolKit.java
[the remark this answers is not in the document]
kirk-microsoft: I prefer tag based logging to levels. My goal is to evict Log4J/SLF4J out of this project.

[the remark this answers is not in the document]
dsgrieve: Kirk and I discussed and this is good for now. We are interested in adding a more complete logging framework c.f. unified logging.

One issue here is the stack trace and impact on performance due to forcing a safe-point. At this point, this is not an issue, but should be addressed at some point.

api/src/main/java/com/microsoft/gctoolkit/message/DataSourceParser.java
dsgrieve: Kirk and I discussed and were astonished that such a great suggestion would come from @johnoliver.

kirk-microsoft (approved):
LGTM....

Issue #203, opened 2022-01-31
Parsing GC Logs can be slow


karianna:
For some logs, parsing is slow.  There are a myriad of factors involved in this, including the order or rules and how they are applied.  The ZGC log attached with permission from AliBaba is an example of this.

[slow_zgc_log.log](https://github.com/microsoft/gctoolkit/files/7969996/slow_zgc_log.log)

dsgrieve:
PR #295 gives about a 28% improvement, but I wouldn't consider this problem solved.

PR #295, opened 2023-06-12
Order keys by mru parse rule

dsgrieve:
Partially addresses issue #203 

Roughly measured, before these changes, a 394M G1GC log file took 111373 ms to parse. With these changes, the same log file took 80211 ms.  A 28% improvement. YMMV.


logesh056:
Het @dsgrieve  can you use hyperScan to find the ruleToApply?
It stores the expression in memory and it validates parseRule fast.

PR #306, opened 2023-09-08
Correct ZGCCycle gcType

loyispa:
Currently ZGCCycle's garbageCollectionTypes is Unknown,  maybe we should correct it.


parser/src/main/java/com/microsoft/gctoolkit/parser/ZGCParser.java
```diff
@@ -334,7 +335,7 @@ public ZGCForwardReference(DateTimeStamp dateTimeStamp, long gcId, GCCause cause
         }
 
         ZGCCycle toZGCCycle(DateTimeStamp endTime) {
-            ZGCCycle cycle = new ZGCCycle(startTimeStamp, gcCause, endTime.minus(startTimeStamp));
+            ZGCCycle cycle = new ZGCCycle(startTimeStamp, GarbageCollectionTypes.ZGCCycle, gcCause, endTime.minus(startTimeStamp));
```
kcpeppe, line 338: GarbageCollectionType is a hangover (now technical debt) from a previous implementation and I had intended on kicking it out of the code base but... it really needs to go. Move over, the constructor *should* set this but allas, it doesn't so the real bug here is in the constructor though this also fixes it so....

PR #309, opened 2023-11-01
Nov 2023 dep updates

karianna:


d3r3kk (approved):
Updates to versions of dependencies and to maven itself. 🚢 it.

d3r3kk:
@karianna Consider updating the GitHub actions in this workflow.

From the result of this PR:

```
"The following actions uses node12 which is deprecated and will be forced to run on node16: actions/checkout@v2, actions/setup-java@v2. For more info: https://github.blog/changelog/2023-06-13-github-actions-all-actions-will-run-on-node16-instead-of-node12-by-default/"
```

PR #314, opened 2023-12-27
Update dependencies and ability to pass in compile version

karianna:


.github/workflows/codeql-analysis.yml
```diff
@@ -66,5 +67,15 @@ jobs:
     #   make bootstrap
     #   make release
 
+    - name: Set up JDK ${{ matrix.java }}
+      uses: actions/setup-java@0ab4596768b603586c0de567f2430c30f5b0d2b0 # v3.13.0
```
dsgrieve, line 71: Why this specific commit? Why not version 4.0? 

This is going to have to change at some point. Is there an issue to track 'update setup-java to <version>'?

.github/workflows/early_access.yml
```diff
@@ -15,8 +15,7 @@ jobs:
     - uses: actions/checkout@b4ffde65f46336ab88eb53be808477a3936bae11 # v4.1.1
 
     - name: Set up JDK 11
-      uses: actions/setup-java@0ab4596768b603586c0de567f2430c30f5b0d2b0 # v3.13.0
-      with:
+      uses: actions/setup-java@387ac29b308b003ca37ba93a6cab5eb57c8f5f93 # v4.0.0
```
dsgrieve, line 18: Strange that this is version 4 and the other is 3.13. Oversight or intention? And why it isn't `@4`

CONTRIBUTING.md
```diff
@@ -16,25 +16,25 @@ or contact [opencode@microsoft.com](mailto:opencode@microsoft.com) with any addi
...
+
 ## Build
 
-The build requires JDK 11, and uses the Maven wrapper (`mvnw`) to help ensure reproducible builds and so we don't force you to change your system Maven install. If you prefer to build with your local Maven installation, make sure the version matches the one in the project's [.mvn/wrapper/maven-wrapper.properties](https://github.com/microsoft/gctoolkit/blob/main/.mvn/wrapper/maven-wrapper.properties) file.
+The build requires JDK 11 (you can also use JDK 17 or 21), and uses the Maven wrapper (`mvnw`) to help ensure reproducible builds and so we don't force you to change your system Maven install. If you prefer to build with your local Maven installation, make sure the version matches the one in the project's [.mvn/wrapper/maven-wrapper.properties](https://github.com/microsoft/gctoolkit/blob/main/.mvn/wrapper/maven-wrapper.properties) file.
 
-* `mvnw clean` - remove build artifacts
-* `mvnw compile` - compile the source code
+* `./mvnw clean` - remove build artifacts
+* `./mvnw compile <11|17|21>` - compile the source code. WARN you must set `11`, `17` or `21` for the release version.
```
dsgrieve, line 25: You don't need `-Dmaven.compiler.release` here? And why can't it default to 21 if maven.compiler.release is not set? 

Maybe what's needed is a github action profile that allows the matrix version. If I'm building this myself, I should be able to do a simple `./mvnw package` without all of the ceremony.

.github/workflows/codeql-analysis.yml
karianna: Should have been 4.0.0 (I've updated that). GH recommends pinning versions to the hash commmit as (something like) v4 is effectively an alias that can silently be changed under the hood. Dependendabot was updated to deal with the hashcommit version numbers and accompanying comment (it will update both)

.github/workflows/early_access.yml
karianna: GH recommends pinning versions to the hash commit as (something like) v4 is effectively an alias that can silently be changed under the hood.

CONTRIBUTING.md
karianna: Good point. In order to support the GH Matrix builds I had to be able to pass in the `maven.compiler.release` property but it would be better with a default somehow.  I'll see if I can figure out how to do that.

Is 21 the default we want? IIRC we have users that have built Java 11 based integrations with this library and we have not used any Java 12+ constructs in the source code.

karianna: OK, I can set the default property in the pom.xml file and it can be overridden by the GH Action matrix build, I've pushed a fix for the pom file and docs

CONTRIBUTING.md
```diff
@@ -16,25 +16,31 @@ or contact [opencode@microsoft.com](mailto:opencode@microsoft.com) with any addi
 ## Submit an Issue
 
 if you wish to contribute to GCToolKit we would kindly ask that you submit an issue to the issue tracker. Doing so will help with the management of the project.
- 
+
 ## Build
 
-The build requires JDK 11, and uses the Maven wrapper (`mvnw`) to help ensure reproducible builds and so we don't force you to change your system Maven install. If you prefer to build with your local Maven installation, make sure the version matches the one in the project's [.mvn/wrapper/maven-wrapper.properties](https://github.com/microsoft/gctoolkit/blob/main/.mvn/wrapper/maven-wrapper.properties) file.
+The build requires JDK 11 and uses the Maven wrapper (`mvnw`) to help ensure reproducible builds and so we don't force you to change your system Maven install. If you prefer to build with your local Maven installation, make sure the version matches the one in the project's [.mvn/wrapper/maven-wrapper.properties](https://github.com/microsoft/gctoolkit/blob/main/.mvn/wrapper/maven-wrapper.properties) file. You can also use JDK 17 or 21 by passing in `-Dmaven.compiler.release=<17|21>` as an extra property to any of the `./mvnw` commands below.
```
dsgrieve, line 22: Probably of no consequence, but if my JDK is 21 and I'm building from the command line, I could give (almost) any release number. 19 for example. I just can't give a release number greater than the version of my JDK.

dsgrieve (approved):
LGTM

CONTRIBUTING.md
karianna: I agree but I think I'd prefer us to only recommend LTS's

PR #315, opened 2023-12-28
Fill region summaries for G1GCFull Event

loyispa:
Seems to have forgotten to fill region summaries for G1GCFull event :)

PR #317, opened 2024-01-02
fix ea release flow

karianna:
- Fill in region summaries for G1GCFull (#315)
- Update dependencies and ability to pass in compile version (#314)
- Fix YAML for ea workflow


dsgrieve (approved):
I'm surprised your IDE didn't red-squigle this. Should we have a YAML validator in the build?

Issue #257, opened 2023-01-17
Improve estimated StartTime calculation in Aggregation


karianna:
* todo: The better way to do this is to calculate the variance in GC frequency and if the gap between 0.000 and
     * the first event exceeds the variance, then the first event would be considered the beginning of the log file.
     * otherwise, the time of the first event should be the time of the first event - the variance.

This removes the 0.25 magic number.

kkdlau:
Hi @karianna, looks like the scope of this issue is well-defined and I think I will be able to handle this - would you like to assign this issue to me 😄 ?

karianna:
> Hi @karianna, looks like the scope of this issue is well-defined and I think I will be able to handle this - would you assign this issue to me 😄 ?

Please go ahead and thank you!

kcpeppe:
merged

PR #360, opened 2024-06-06
Estimate the start time by variance

kkdlau:
This PR is for fixing #257

Below is a summary of the changes made in this pull request:
- New `OnlineCalculator` classes - `WelfordVarianceCalculator` and `OnlineStatisticsCalculator`
  - This two classes allows to compute the mean and variance on-the-fly, without caching the dataset in memory
- Test cases for `OnlineCalculator` classes
- Update the way of computing estimated start time

Regarding the estimated start time, I used a slightly different approach than suggested in #257
- GC frequency is calculated as the time gap between two consecutive GC events.
  - For example, if event A happens at 2s and event B happens at 2.3s, the GC frequency is 2.3s - 2s = 0.3s.
- If `timeOfFirstEvent` does not have time stamp, the method will return `date stamp - SD of GC frequency`
- If `timeOfFirstEvent` has time stamp
 - it will first compute  `time stamp - SD of GC frequency`
   - if the value is greater than 0, then return value directly
   - otherwise return `timeOfFirstEvent`, since we cannot have negative time stamp

Please review the code and provide any comments or feedback you may have. I would also appreciate clarification on the specific points mentioned above!

### DIFF
--- a/IT/src/test/java/com/microsoft/gctoolkit/integration/CaptureJVMTerminationEventTest.java
+++ b/IT/src/test/java/com/microsoft/gctoolkit/integration/CaptureJVMTerminationEventTest.java
@@ -66,11 +66,11 @@ public class CaptureJVMTerminationEventTest {
             fail(e.getMessage());
         }
         Assertions.assertEquals( 2.193d, terminationAggregation.getStartTime().getTimeStamp(),0.001d, "Time of first event");
-        Assertions.assertEquals( 608800.088d, heapOccupancyAfterCollectionSummary.estimatedRuntime(),0.001d, "Runtime duration");
+        Assertions.assertEquals( 608797.895d, heapOccupancyAfterCollectionSummary.estimatedRuntime(),0.001d, "Runtime duration");
 
     }
    
-   @Aggregates({EventSource.CMS_PREUNIFIED,EventSource.CMS_UNIFIED,EventSource.G1GC,EventSource.GENERATIONAL,EventSource.JVM,EventSource.SHENANDOAH, EventSource.TENURED,EventSource.ZGC})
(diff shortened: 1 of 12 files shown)


kkdlau:

kcpeppe (approved):

karianna:
Hi @kkdlau - @kcpeppe sees this as a good change :-), are you able to go ahead and also update the tests to match?

kkdlau:
> Hi @kkdlau - @kcpeppe sees this as a good change :-), are you able to go ahead and also update the tests to match?

definitely! Will push the change later

dsgrieve (approved):

IT/src/test/java/com/microsoft/gctoolkit/integration/CaptureJVMTerminationEventTest.java
```diff
@@ -66,11 +66,11 @@ public void analyze(String gcLogFile) {
             fail(e.getMessage());
         }
         Assertions.assertEquals( 2.193d, terminationAggregation.getStartTime().getTimeStamp(),0.001d, "Time of first event");
-        Assertions.assertEquals( 608800.088d, heapOccupancyAfterCollectionSummary.estimatedRuntime(),0.001d, "Runtime duration");
+        Assertions.assertEquals( 608797.895d, heapOccupancyAfterCollectionSummary.estimatedRuntime(),0.001d, "Runtime duration");
 
     }
    
-   @Aggregates({EventSource.CMS_PREUNIFIED,EventSource.CMS_UNIFIED,EventSource.G1GC,EventSource.GENERATIONAL,EventSource.JVM,EventSource.SHENANDOAH, EventSource.TENURED,EventSource.ZGC})
+   @Aggregates({EventSource.GENERATIONAL,EventSource.CMS_UNIFIED,EventSource.G1GC,EventSource.GENERATIONAL,EventSource.JVM,EventSource.SHENANDOAH, EventSource.TENURED,EventSource.ZGC})
```
karianna, line 73: This does seem incorrect? It's a duplicate?

karianna (changes requested):

Issue #352, opened 2024-05-01
Generational parser does not seem to handle GC Cause included in young event.


Bluetopia:
**Describe the bug**
When processing a line like:
```5.374: [GC (Allocation Failure) [PSYoungGen: 262144K->28340K(305664K)] 262144K->28364K(1005056K), 0.0146665 secs] [Times: user=0.08 sys=0.00, real=0.01 secs]```
GenerationalHeapParser.psYoungGen(GCLogTrace, String) is called and the offset (5) appears to not take into account the cause, resulting in:

```
May 01, 2024 10:03:19 AM com.microsoft.gctoolkit.parser.GCLogTrace getOccupancyBeforeAfterWithMemoryPoolSizeSummary
WARNING: Unable to calculate generational memory pool summary.
vert.x-eventloop-thread-0, not implemented: 5.374: [GC (Allocation Failure) [PSYoungGen: 262144K->28340K(305664K)] 262144K->28364K(1005056K), 0.0146665 secs]2024-05-01 10:03:19,806 [vert.x-eventloop-thread-0] INFO STDOUT: vert.x-eventloop-thread-0, not implemented: 5.374: [GC (Allocation Failure) [PSYoungGen: 262144K->28340K(305664K)] 262144K->28364K(1005056K), 0.0146665 secs]
```

with the following groups:

```
1: 5.374
2: null
3: null
4: 5.374
5: null
6: (Allocation Failure)
7: PSYoungGen
8: 262144
9: K
10: 28340
11: K
12: 305664
13: K
14: 262144
15: K
16: 28364
17: K
18: 1005056
19: K
20: 0.0146665
```

getOccupancyBeforeAfterWithMemoryPoolSizeSummary tries to convert null (5), "PSYoungGen" (7), and "K" to KByte values, which generates a NumberFormatException, and causing the notYetImplemented() logging.

**To Reproduce**
* Attempt to load the attached file
[sample.zip](https://github.com/microsoft/gctoolkit/files/15178111/sample.zip)

**Expected behavior**
GCToolkit is able to properly parse the GC event. 

**Desktop (please complete the following information):**
 - OS: Windows 11
 - Java 11 (Amazon Corretto)
 - GCToolkit tip (April 30, 2024)

**Additional context**
This is an older log file from Java 8/CMS that likely won't come up as frequently anymore, so it may be a lower priority.  It's not clear if this should be a feature request instead due to the fact this triggers "Not yet implemented" logging.

[...]

PR #362, opened 2024-06-11
Fix for gctoolkit #352 (Generational Heap Parser fails to recognize parallel gc lines)

jlittle-ptc:
The bug was caused by mismatched indices in the regex that matches a Parallel collection.  The indices of two methods (YoungGen and Failure) in GenerationalHeapParser have been updated to correctly match the expected index so that the lines no longer generate a "not implemented" error in debug mode. 

I also reviewed the test suite to determine why this issue wasn't caught, and found that there was no test case that explicitly tested the parsing of Parallel GC lines, so I've added an additional test case for PSYoungGen that fails before the changes, and passes afterwards.  

I was unable to add a similar test case for Failure using the current strategy of parse and check for events as the failure line appears to be ignored. 

There is a chance that further fixes are required for less-frequently logged lines, but for any sample data I have available, the files now parse successfully.


jlittle-ptc:
@microsoft-github-policy-service agree [company="PTC"]

karianna (approved):
LGTM - just @kcpeppe to review

dsgrieve (approved):

Issue #364, opened 2024-06-19
Generational  ZGC not supported


loyispa:
The ZGC log pattern seems to have changed since jdk21, do you have a plan to support it?

Known changes for **CYCLE_START**:

before:

```
[20.134s][info ][gc          ] GC(1) Garbage Collection (System.gc()) 22M(1%)->24M(1%)
```

after:
```
[251.570s][info][gc          ] GC(5) Major Collection (Proactive) 684M(23%)->346M(12%) 0.473s
```

karianna:
Yes, Pull Requests are also welcome here though (it's an ever moving target!)

kcpeppe:
Hi, I have work on genZ underway. I'm willing to push the work in my local branch if there is interest

loyispa:
> Hi, I have work on genZ underway. I'm willing to push the work in my local branch if there is interest

@kcpeppe  I am sure interested.  But sorry for my naive question, why is it a local branch not the official repository?

loyispa:
I also  have a preliminary implementation regarding genz, and can meet my own needs ,don't know if it should be submited.

karianna:
You should both push branches as part of a PR request and let's have a look at them!

kcpeppe:
Yes, please push and let's see if we can integrate them.

AlinaHovakimyan:
Hi everyone, may I ask when the next version is planned to be released? We are also looking forward for the generational ZGC support and I see that the latest version available in maven central is from a year ago.

PR #365, opened 2024-06-26
Support generational ZGC

loyispa:
Hi, here is a enhance for Generational ZGC , towards https://github.com/microsoft/gctoolkit/issues/364

major changes:
1. Some events(PauseMark,ConcurrentMark, etc.) occur twice in a single Cycle,  using their first timestamp.
2. Sum the event durations and memory occupancies (live/garbage/allocated) for the young generation and old generation.

### DIFF
--- a/parser/src/main/java/com/microsoft/gctoolkit/parser/ZGCParser.java
+++ b/parser/src/main/java/com/microsoft/gctoolkit/parser/ZGCParser.java
@@ -59,6 +59,7 @@ public class ZGCParser extends UnifiedGCLogParser implements ZGCPatterns {
         parseRules.put(PAUSE_PHASE, this::pausePhase);
         parseRules.put(CONCURRENT_PHASE, this::concurrentPhase);
         parseRules.put(LOAD, this::load);
+        parseRules.put(GENERATION_LOAD, this::load);
         parseRules.put(MMU, this::mmu);
         parseRules.put(MARK_SUMMARY, this::markSummary);
         parseRules.put(RELOCATION_SUMMARY, this::relocationSummary);
@@ -66,6 +67,7 @@ public class ZGCParser extends UnifiedGCLogParser implements ZGCPatterns {
         parseRules.put(METASPACE, this::metaspace);
         parseRules.put(REFERENCE_PROCESSING, this::referenceProcessing);
         parseRules.put(CAPACITY, this::capacity);
+        parseRules.put(GENERATIONAL_MEMORY_STATISTICS, this::memoryStatistics);
         parseRules.put(MEMORY_TABLE_ENTRY_SIZE, this::sizeEntry);
         parseRules.put(MEMORY_TABLE_ENTRY_OCCUPANCY, this::occupancyEntry);
         parseRules.put(MEMORY_TABLE_ENTRY_RECLAIMED, this::reclaimed);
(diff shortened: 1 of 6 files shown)


parser/src/main/java/com/microsoft/gctoolkit/parser/ZGCParser.java
```diff
@@ -323,6 +340,7 @@ private class ZGCForwardReference {
         private ReclaimSummary reclaimed;
         private ReclaimSummary memorySummary;
         private ZGCMetaspaceSummary metaspace;
+        private int memoryScope; // 0: default all heap; 1: young generation; 2: old generation
```
karianna, line 343: I'm thinking enum here?

parser/src/main/java/com/microsoft/gctoolkit/parser/ZGCParser.java
```diff
@@ -363,107 +381,150 @@ ZGCCycle toZGCCycle(DateTimeStamp endTime) {
...
         public void setPauseMarkEndStart(DateTimeStamp pauseMarkEndStart) {
+            if (this.pauseMarkEndStart != null) {
+                return;
+            }
             this.pauseMarkEndStart = pauseMarkEndStart;
         }
 
         public void setPauseMarkEndDuration(double pauseMarkEndDuration) {
-            this.pauseMarkEndDuration = pauseMarkEndDuration;
+            this.pauseMarkEndDuration += pauseMarkEndDuration;
```
karianna, line 402: Is this GenZGC vs ZGC reasoning or fixing what you feel is a bug?

karianna: Same with the other += changes.

karianna (commented):
Couple of early comments, I'll leave the overall review to @kcpeppe et al.

parser/src/main/java/com/microsoft/gctoolkit/parser/ZGCParser.java
loyispa: Due to genZ dividing pauseMarkEndDuration into young and old duration, so sum them equals the original meaning.
```
[4.120s][info][gc,phases   ] GC(0) Y: Pause Mark Start (Major) 0.066ms
......
[4.155s][info][gc,phases   ] GC(0) Y: Pause Mark End 0.016ms
......
[4.182s][info][gc,phases   ] GC(0) O: Pause Mark End 0.014ms
......
```

parser/src/main/java/com/microsoft/gctoolkit/parser/ZGCParser.java
loyispa: Yes, it's much better than a magic number.

kcpeppe:
Currently a single ZGC cycle is represented by a single event. For genZ I was looking at 2 events. I've thought of having an event per phase but I was concerned that this might lead to some downstream confusion with the analytics. Also, as you've mentioned, repeated phases per cycle is an issue. This may become even more confused depending on the log levels used by different projects. So, I'm leaning towards the single event per cycle but I'm not stuck on it. Are there reasons to create event per phase within a cycle that I'm missing?

loyispa:
I'm inclined to think you're right to be concerned. Creating events per phase seems to break the "consensus" that each GCID represents a separate gc event. It may cause confusion in the counting of gc events.

loyispa:
@kcpeppe   Do you have time to take a look at this PR?  Or if you have new plans for the genZ,  please let me know :)

kcpeppe:
I think I'd like to see the events collapsed into one for a young and one for a tenured collection.

kcpeppe:
sorry for dribbling thoughts here but I'm thinking that genZ should be it's own parser.

kcpeppe:
This PR has caused me to reverse my thinking about how to integrate genZ into GCToolKit. I'd like to make a couple of mods and then merge.

PR #408, opened 2024-12-20
Unified Serial GC correction

kcpeppe:
Bunch of fixes but the important one is the fix to produce System GC full events.


parser/src/main/java/com/microsoft/gctoolkit/parser/UnifiedGenerationalParser.java
```diff
@@ -481,12 +483,21 @@ private FullGC fillOutFullGC(FullGC event, GenerationalForwardReference values)
     }
 
     private FullGC buildFullGC(GenerationalForwardReference forwardReference) {
+        FullGC gc;
         switch (forwardReference.getGarbageCollectionType()) {
             case PSFull:
-                return fillOutFullGC(new PSFullGC(forwardReference.getStartTime(), forwardReference.getGCCause(), forwardReference.getDuration()), forwardReference);
+                if ( forwardReference.getGCCause().equals(GCCause.JAVA_LANG_SYSTEM))
```
karianna, line 489: It feels like some of this is DRY / could be squashed down, on the other hand it's readable, so that wins!

parser/src/test/java/com/microsoft/gctoolkit/parser/unittests/UnifiedGenerationalParserTest.java
```diff
@@ -65,4 +65,32 @@ public void testCMSLogsMissingTimeStamps() {
...
+            "factorization-serialgc-tip.log"
+    };
+
+    private static final int[] serialNumberOfDifferentCollectors = {
+            2,
+    };
+
+    private static final int[][] serialCounts = {
+            //  0,   1,    2,     3,     4,     5,     6,    7,    8,     9,      10,     11,      12, 13, 14, 15, 16, 17
+            {   0,  13,    0,     0,     0,     0,     0,    0,    0,     0,       4,      0,       0,  0,  0,  0,  0,  0},
```
karianna, line 94: A new developer coming into this will struggle, can we add a comment stating what this array effectively tests?

karianna (approved):

dsgrieve (approved):

PR #414, opened 2025-02-20
ZGC Generational Support

j-bahr:
This patch adds support for generational ZGC and maintains support for legacy (non generational ZGC). There are a number of potential API breaking changes for consumers of GC events. List of changes and modifications are listed below.

1. Introduce new GCEvent types, FullZGCCycle (legacy), MajorZGCCycle and MinorZGCCycle representing the respective types.
2. GCEvents Full/Major/Minor GC Cycles hold reference to internal ZGCCycle, a delegate in the case of Full (legacy) cycle, young and for Major cycle and simply a young for Minor cycle. All cycle types full/young/old, have mostly similar overlapping fields, which allows us to have a single data class for all cycle types. Cycles themselves have a few fields that are specific to entire cycle and are held in the top level object.

```
┌─────────────┐
│    Full     │
│             │      ┌────────┐
│  delegate ──┼─────▶│ZGCCycle│
│             │      └────────┘
└─────────────┘
┌─────────────┐
│    Major    │
│             │      ┌────────┐
│    young ───┼─────▶│ZGCCycle│
│             │      ├────────┤
│     old  ───┼─────▶│ZGCCycle│
└─────────────┘      └────────┘
┌─────────────┐
│    Minor    │
│             │     ┌────────┐
│    young ───┼────▶│ZGCCycle│
│             │     └────────┘
└─────────────┘

```

3.  Add stronger typing on the heap summary report. Live, Garbage, Allocated, Reclaimed, Promoted and Compacted carry information for a different stages of the GC. Each holds a different number of fields. Having a single summary with primitive `double` doesn't make sense as you can't differentiate between 0 and not present. We could use optional or boxed type but it seemed more direct to declare a type. In future java versions these could simply be Record classes.
4. Added concept of phase to the ZGC Parsers. The phase represents whether the GC cycle is in Major young/old, Minor young or Full phase (legacy). This maps to the GC output of `Y`, `y`, `O` prefix on the log line. We key off this to make a determination about which heap statistics we are writing to. Phase is determined by an optional capture group at the head of all log lines.
5. Introduced new hierarchical forward reference type to allow it to cary more contextual information about either young or old generation. Major obviously has two references, one for old and one for young. Minor and full GC's cary a single forward ref corresponding to their cycle.
6. Added new ZGC generational test for both Minor and Major cycle types ensuring that both are parsed correctly.

### DIFF
--- a/api/src/main/java/com/microsoft/gctoolkit/event/zgc/ZGCCollectionType.java
+++ b/api/src/main/java/com/microsoft/gctoolkit/event/zgc/ZGCCollectionType.java
@@ -1,0 +1,22 @@
+package com.microsoft.gctoolkit.event.zgc;
+
+public enum ZGCCollectionType {
+    FULL("Garbage"), // Legacy ZGC
+    MINOR("Minor"),
+    MAJOR("Major");
+
+    private final String collectionLabel;
+
+    ZGCCollectionType(String collectionLabel) {
+        this.collectionLabel = collectionLabel;
+    }
+
+    public static ZGCCollectionType get(String label) {
+        for (ZGCCollectionType status : ZGCCollectionType.values()) {
+            if (status.collectionLabel.equalsIgnoreCase(label.trim())) {
+                return status;
+            }
+        }
+        throw new IllegalArgumentException("No matching ZGCCollectionType found for: " + label);
+    }
+}
--- a/api/src/main/java/com/microsoft/gctoolkit/event/zgc/ZGCGarbageSummary.java
+++ b/api/src/main/java/com/microsoft/gctoolkit/event/zgc/ZGCGarbageSummary.java
@@ -1,0 +1,27 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+package com.microsoft.gctoolkit.event.zgc;
+
+public class ZGCGarbageSummary {
+    private final long markEnd;
+    private final long relocateStart;
+    private final long relocateEnd;
+
+    public ZGCGarbageSummary(long markEnd, long relocateStart, long relocateEnd) {
+        this.markEnd = markEnd;
+        this.relocateStart = relocateStart;
+        this.relocateEnd = relocateEnd;
+    }
+
+    public long getMarkEnd() {
+        return markEnd;
+    }
+
+    public long getRelocateStart() {
+        return relocateStart;
+    }
+
+    public long getRelocateEnd() {
+        return relocateEnd;
+    }
+}
--- a/api/src/main/java/com/microsoft/gctoolkit/event/zgc/ZGCMemorySummary.java
+++ b/api/src/main/java/com/microsoft/gctoolkit/event/zgc/ZGCMemorySummary.java
@@ -1,0 +1,20 @@
+package com.microsoft.gctoolkit.event.zgc;
+
+public class ZGCMemorySummary {
+    private final long start;
+    private final long end;
+
+    public ZGCMemorySummary(long start, long end) {
+        this.start = start;
+        this.end = end;
+    }
+
+    public long getStart() {
+        return start;
+    }
+
+    public long getEnd() {
+        return end;
+    }
+
+}
--- a/api/src/main/java/com/microsoft/gctoolkit/event/zgc/ZGCPhase.java
+++ b/api/src/main/java/com/microsoft/gctoolkit/event/zgc/ZGCPhase.java
@@ -1,0 +1,27 @@
+package com.microsoft.gctoolkit.event.zgc;
+
+public enum ZGCPhase {
+    FULL(""),
+    MAJOR_YOUNG("Y"),
+    MAJOR_OLD("O"),
+    MINOR_YOUNG("y");
+
+    private final String phase;
+
+    ZGCPhase(String s) {
+        this.phase = s;
+    }
+
+    public String getPhase() {
+        return phase;
+    }
+
+    public static ZGCPhase get(String label) {
+        for (ZGCPhase zgcPhase : ZGCPhase.values()) {
+            if (zgcPhase.getPhase().equals(label.trim())) {
+                return zgcPhase;
+            }
+        }
+        throw new IllegalArgumentException("No matching ZGCPhase found for: " + label);
+    }
+}
--- a/parser/src/main/java/com/microsoft/gctoolkit/parser/GCLogTrace.java
+++ b/parser/src/main/java/com/microsoft/gctoolkit/parser/GCLogTrace.java
@@ -10,10 +10,14 @@ import com.microsoft.gctoolkit.event.UnifiedCountSummary;
 import com.microsoft.gctoolkit.event.UnifiedStatisticalSummary;
 import com.microsoft.gctoolkit.event.jvm.MetaspaceRecord;
 import com.microsoft.gctoolkit.event.jvm.PermGenSummary;
+import com.microsoft.gctoolkit.event.zgc.ZGCCollectionType;
+import com.microsoft.gctoolkit.event.zgc.ZGCPhase;
 
 import java.util.logging.Level;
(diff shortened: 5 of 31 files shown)


adolfojunior:
👍

karianna:
Thanks for this PR!  It may take us a day or two to pick through it, but it's appreciated.

karianna:
On a side note: @kcpeppe & @dsgrieve - Not sure if our LICENSE file covers everything or whether we need to do a separate PR to add the minimal license header to all of the files that are missing it.

parser/src/main/java/com/microsoft/gctoolkit/parser/GCLogTrace.java
```diff
@@ -289,4 +302,13 @@ public void notYetImplemented() {
         System.out.println("-----------------------------------------");
         //}
     }
+
+    public ZGCPhase getZCollectionPhase() {
```
kcpeppe, line 306: A GCLogTrace is a convenience wrapper over the Matcher. It contains convenience methods to extract data from the matcher. It is expected that the parser has the context to know which methods to call and which groups need to be extracted. Thus ZGCPhase is either known and extracted by the parser at that time *or* ZGCPhase isn't present in the log line (matcher) and the parser, again, should know this and not call this method. Either way, ZGCPhase should be consistent and there should be no need for downstream exception handling. This is something that was noted in ZGCPhase that an illegal argument was being thrown which likely is unnecessary.

parser/src/main/java/com/microsoft/gctoolkit/parser/GCLogTrace.java
```diff
@@ -85,6 +89,15 @@ public double getMilliseconds(int index) {
         return getDoubleGroup(index);
     }
 
+    /**
```
kcpeppe, line 92: The parser should know this.. and it should be picked up in testing

api/src/main/java/com/microsoft/gctoolkit/event/zgc/ZGCCollectionType.java
```diff
@@ -0,0 +1,22 @@
...
+        this.collectionLabel = collectionLabel;
+    }
+
+    public static ZGCCollectionType get(String label) {
+        for (ZGCCollectionType status : ZGCCollectionType.values()) {
+            if (status.collectionLabel.equalsIgnoreCase(label.trim())) {
+                return status;
+            }
+        }
+        throw new IllegalArgumentException("No matching ZGCCollectionType found for: " + label);
```
kcpeppe, line 20: We should never get here.

api/src/main/java/com/microsoft/gctoolkit/event/zgc/ZGCGarbageSummary.java
```diff
@@ -0,0 +1,27 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+package com.microsoft.gctoolkit.event.zgc;
+
+public class ZGCGarbageSummary {
```
kcpeppe, line 5: are these values heap occupancy before or after the collection?

api/src/main/java/com/microsoft/gctoolkit/event/zgc/ZGCMemorySummary.java
```diff
@@ -0,0 +1,20 @@
+package com.microsoft.gctoolkit.event.zgc;
+
+public class ZGCMemorySummary {
+    private final long start;
+    private final long end;
+
+    public ZGCMemorySummary(long start, long end) {
+        this.start = start;
+        this.end = end;
```
kcpeppe, line 9: start and end are ambiguous, can these names be more descriptive?

api/src/main/java/com/microsoft/gctoolkit/event/zgc/ZGCPhase.java
```diff
@@ -0,0 +1,27 @@
...
+        return phase;
+    }
+
+    public static ZGCPhase get(String label) {
+        for (ZGCPhase zgcPhase : ZGCPhase.values()) {
+            if (zgcPhase.getPhase().equals(label.trim())) {
+                return zgcPhase;
+            }
+        }
+        throw new IllegalArgumentException("No matching ZGCPhase found for: " + label);
```
kcpeppe, line 25: should never get here

parser/src/main/java/com/microsoft/gctoolkit/parser/GenericTokens.java
```diff
@@ -20,6 +20,7 @@ public interface GenericTokens {
     //Time
     String TIME = "(-?" + REAL_NUMBER + ")";
     String DURATION_MS = TIME + "\\s?ms";
+    String DURATION_S = TIME + "\\s?s";
     String INT_DURATION_MS = INTEGER + "ms";
     //0.0700188
     String PAUSE_TIME = TIME + "\\s?(?:secs?|ms)";
```
kcpeppe, line 26: I would have added the s to the PAUSE_TIME pattern but that is just me.

parser/src/test/java/com/microsoft/gctoolkit/parser/ShenandoahParserTest.java
```diff
@@ -4,6 +4,9 @@
 
 import com.microsoft.gctoolkit.event.jvm.JVMEvent;
 import com.microsoft.gctoolkit.event.shenandoah.ShenandoahCycle;
+import com.microsoft.gctoolkit.event.zgc.OccupancySummary;
```
kcpeppe, line 7: are these changes intentional or should they be reverted? The added methods are never called?

parser/src/main/java/com/microsoft/gctoolkit/parser/ZGCParser.java
```diff
@@ -196,117 +320,316 @@ private void nMethods(GCLogTrace trace, String s) {
...
         super.publish(ChannelName.ZGC_PARSER_OUTBOX, event);
         forwardReference = null;
     }
 
-    private class ZGCForwardReference {
+    private interface ZFwdRef {
+        GCEvent getGCEVent(DateTimeStamp endTime);
+    }
+
+    private static class ZGCMinorForwardReference implements ZFwdRef {
```
kcpeppe, line 519: would prefer if this was ZForwardReference instead of ZFwdRef

kcpeppe (approved):
Thank you for the very high quality patch. We will release this ASAP, likely tomorrow (Feb 26). Please create a PR for any comments you would like to address.

dsgrieve:
@j-bahr Thank you for your contribution which is now in gctoolkit 3.5.0 on Maven central.

j-bahr:
@kcpeppe, Thanks for the review I will incorporate feedback into another PR.

api/src/main/java/com/microsoft/gctoolkit/event/zgc/ZGCMemorySummary.java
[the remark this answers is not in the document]
j-bahr: I think this makes sense. The memory summary was meant to capture the closing line on any of the cycles eg

```
GC(40) Garbage Collection (Proactive) 698M(2%)->262M(1%)
``` 
where the occupancy was 698M before and 262M after. 

There are other other log lines that convey the capacity of the heap which could be combined to form a complete summary. I'm happy to attempt to implement this.
```
GC(40) Min Capacity: 8M(0%)
GC(40) Max Capacity: 30718M(100%)
GC(40) Soft Max Capacity: 30718M(100%)
```

parser/src/test/java/com/microsoft/gctoolkit/parser/ShenandoahParserTest.java
j-bahr: I think this was a rebase error. I'll revert.

PR #417, opened 2025-02-28
ZGC Fixes from #414

j-bahr:
Minor fixes from feedback in #414

- Fix enum types to avoid having to throw IllegalArgumentException
- Fix Memory summary to have more descriptive variable names
- ZGCPhase now handles the regex null case correctly for Full (legacy) phase types
- Use pause time to collect optional GC duration
- Rename ZFwdRef -> ZForwardReference
- Remove unused methods in ShenandoahParserTest.java

### DIFF
--- a/api/src/main/java/com/microsoft/gctoolkit/event/zgc/ZGCCollectionType.java
+++ b/api/src/main/java/com/microsoft/gctoolkit/event/zgc/ZGCCollectionType.java
@@ -1,5 +1,8 @@
 package com.microsoft.gctoolkit.event.zgc;
 
+import java.util.Arrays;
+import java.util.Objects;
+
 public enum ZGCCollectionType {
     FULL("Garbage"), // Legacy ZGC
     MINOR("Minor"),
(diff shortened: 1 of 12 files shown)


api/src/main/java/com/microsoft/gctoolkit/event/zgc/ZGCCollectionType.java
```diff
@@ -12,11 +15,9 @@ public enum ZGCCollectionType {
     }
 
     public static ZGCCollectionType get(String label) {
-        for (ZGCCollectionType status : ZGCCollectionType.values()) {
-            if (status.collectionLabel.equalsIgnoreCase(label.trim())) {
-                return status;
-            }
-        }
-        throw new IllegalArgumentException("No matching ZGCCollectionType found for: " + label);
+        return Arrays.stream(ZGCCollectionType.class.getEnumConstants())
```
kcpeppe, line 18: I'll accept this but... what do you think about introducing "Unknown" and return that instead of returning null?

api/src/main/java/com/microsoft/gctoolkit/event/zgc/ZGCPhase.java
```diff
@@ -17,11 +20,9 @@ public String getPhase() {
...
-        for (ZGCPhase zgcPhase : ZGCPhase.values()) {
-            if (zgcPhase.getPhase().equals(label.trim())) {
-                return zgcPhase;
-            }
-        }
-        throw new IllegalArgumentException("No matching ZGCPhase found for: " + label);
+        return Arrays.stream(ZGCPhase.class.getEnumConstants())
+                .filter(phase -> Objects.equals(phase.getPhase(), label))
+                .findFirst()
+                .orElse(null);
```
kcpeppe, line 26: same as previous comment about Unknown

PR #419, opened 2025-02-28
Change ZGC start tag

### LINKED ISSUE #382: Unrecognised ZGC file
gctoolkit is not being able to recognize ZGC file generated with non-debug mode. Issue is because it indicates ZGC file by looking for "Initializing The Z Garbage Collector" line, which is not there if you just generate zgc log by bin/java -XX:+UseZGC , but if I generate it on debug mode, then all is good: bin/java -XX:+UseZGC -Xlog:gc*,gc+director=debug:file=cms_generated.log:time,level,tags. The line that exists in both files and I guess could be used is "Using The Z Garbage Collector"

j-bahr:
This change allows us to detect a ZGC log when only using `-Xlog:gc`. The old tag required `gc+init=info`. This should improve log detection. This should help fixes #382

Old detection using:
```
[info][gc,init] Initializing The Z Garbage Collector
```

New detection using:
```
[info][gc     ] Using The Z Garbage Collector
```


karianna:
@kcpeppe I know you wanted to have a min set of logging flags that we would support - flagging this for your review

kcpeppe:
I like this change a lot. The more we can do with less info the better IMO. @karianna

Issue #421, opened 2025-03-04
Direct logging to stdout


AlinaHovakimyan:
gctoolkit does direct logging via System.out.println in few places, most importantly in GCLogTrace.java (notYetImplemented function) in non-debug mode. Can it be please enabled in debug mode only or add some common word to all of the logs, so we can filter out this logs from the console output ?

karianna:
PR would be welcome for this!

AlinaHovakimyan:
opened PR:  [#422](https://github.com/microsoft/gctoolkit/pull/422)

PR #422, opened 2025-03-06
Move stdout direct logging under debug mode

AlinaHovakimyan:
This direct logging via System.out.println in non-debug mode causes noice in the console output for the tool using gctoolkit. Discussion: [421](https://github.com/microsoft/gctoolkit/issues/421)

### DIFF
--- a/parser/src/main/java/com/microsoft/gctoolkit/parser/GCLogTrace.java
+++ b/parser/src/main/java/com/microsoft/gctoolkit/parser/GCLogTrace.java
@@ -2,6 +2,7 @@
 // Licensed under the MIT License.
 package com.microsoft.gctoolkit.parser;
 
+import com.microsoft.gctoolkit.GCToolKit;
 import com.microsoft.gctoolkit.event.GCCause;
 import com.microsoft.gctoolkit.event.GCCauses;
 import com.microsoft.gctoolkit.event.MemoryPoolSummary;
@@ -28,7 +29,6 @@ public class GCLogTrace extends AbstractLogTrace {
     private static final Logger LOGGER = Logger.getLogger(GCLogTrace.class.getName());
 
     private final boolean gcCauseDebugging = Boolean.getBoolean("microsoft.debug.gccause");
-    private final boolean debugging = Boolean.getBoolean("microsoft.debug");
 
     public GCLogTrace(Matcher matcher) {
         super(matcher);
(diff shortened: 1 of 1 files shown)


parser/src/main/java/com/microsoft/gctoolkit/parser/GCLogTrace.java
```diff
@@ -28,7 +28,7 @@ public class GCLogTrace extends AbstractLogTrace {
     private static final Logger LOGGER = Logger.getLogger(GCLogTrace.class.getName());
 
     private final boolean gcCauseDebugging = Boolean.getBoolean("microsoft.debug.gccause");
-    // private final boolean debugging = Boolean.getBoolean("microsoft.debug");
+    private final boolean debugging = Boolean.getBoolean("microsoft.debug");
```
karianna, line 31: @kcpeppe I suspect we want ot change this to gctoolkit.debug as opposed to microsoft. -thoughts?

dsgrieve: See also https://github.com/microsoft/gctoolkit/blob/d0f19a9c2eab468ad91820bd139d7a6fcdf00c48/api/src/main/java/com/microsoft/gctoolkit/GCToolKit.java#L42-L74

I'd prefer to see debug flags in the GCToolKit class and all debugging go through the GCToolKit.LOG_DEBUG_MESSAGE method. The code below would be something like:
```java
GCToolKit.LOG_DEBUG_MESSAGE(() -> {
    StringBuilder debugMessage = new StringBuilder();
    debugMessage.append(threadName).append(", not implemented: ").append(getGroup(0)).append(System.lineSeparator());
    for (int i = 1; i < groupCount() + 1; i++) {
        debugMessage.append(i).append(": ").append(getGroup(i)).append(System.lineSeparator());
    }
    debugMessage.append("-----------------------------------------");
}
```
The lambda is not evaluated if gctoolkit.debug is not enabled.

AlinaHovakimyan:

AlinaHovakimyan:
Hi gctoolkit team: @karianna , @dsgrieve, @kcpeppe 
Is there any queries related to this PR or it look good to be merged?

dsgrieve (approved):

dsgrieve:
Thanks @AlinaHovakimyan!

PR #423, opened 2025-03-10
Ensure FullZGCCycle has all delegate methods

j-bahr:
FullZGCCycle delegates it's accessors to underlying ZGCCycle class. The FullZGCCycle was missing some delegate accessor methods missed during a previous refactor. This change ensures that all "getter" methods are exposed from the FullZGCCycle. These were regenerated automatically by intellij hence the reformatting and reordering of methods.


karianna (approved):

dsgrieve (approved):

dsgrieve:
Thanks @j-bahr!

Issue #372, opened 2024-08-07
YoungDetails event not published for GC log collected with -Xlog:gc (no *)


jlittle-ptc:
**Describe the bug**
The attached file seems to be generated by using the options -Xlog:gc (not -Xlog:gc*).  Without the *, only the events tagged "gc" are written out.
```
[6.837s][info][gc] GC(3) Pause Young (Normal) (G1 Evacuation Pause) 79M->27M(1024M) 2.929ms
[7.292s][info][gc] GC(4) Pause Young (Normal) (G1 Evacuation Pause) 74M->28M(1024M) 3.038ms
[7.619s][info][gc] GC(5) Pause Young (Concurrent Start) (Metadata GC Threshold) 46M->30M(1024M) 4.745ms
[7.620s][info][gc] GC(6) Concurrent Cycle
[7.629s][info][gc] GC(6) Pause Remark 31M->31M(1024M) 2.697ms
[7.631s][info][gc] GC(6) Pause Cleanup 31M->31M(1024M) 0.073ms
[7.632s][info][gc] GC(6) Concurrent Cycle 11.893ms
```

While I originally mistakenly thought there was a mismatch on the rule, Events 3 and 4 do successfully match the `YOUNG_DETAILS` parse rule.  However, I suspect the problem is that the forwardReference is populated by `UnifiedG1GCParser.youngDetails()`, there is nothing in the log file to cause the forwardReference to be published.  

I confirmed this behaviour by setting up breakpoints on all publish*() methods in the class.  

The output for -Xlog:gc* is more verbose and contains a line that triggers the publish:
```
[2024-08-07T13:31:02.129-0400][gc,start     ] GC(0) Pause Young (Normal) (G1 Evacuation Pause)
[2024-08-07T13:31:02.130-0400][gc,task      ] GC(0) Using 10 workers of 10 for evacuation
[2024-08-07T13:31:02.133-0400][gc,phases    ] GC(0)   Pre Evacuate Collection Set: 0.0ms
[2024-08-07T13:31:02.133-0400][gc,phases    ] GC(0)   Evacuate Collection Set: 2.5ms
[2024-08-07T13:31:02.133-0400][gc,phases    ] GC(0)   Post Evacuate Collection Set: 0.2ms
[2024-08-07T13:31:02.133-0400][gc,phases    ] GC(0)   Other: 1.4ms
[2024-08-07T13:31:02.135-0400][gc,heap      ] GC(0) Eden regions: 12->0(10)
[...]

PR #430, opened 2025-04-21
Fix for Gctoolkit #372 - YoungDetails event not published for GC log collected with -Xlog:gc (no *)

jlittle-ptc:
**Approach:**

Both files that are generated using -Xlog:gc and -Xlog:gc* use the same general format for the log line of interest:
-Xlog:gc: `[6.837s][info][gc] GC(3) Pause Young (Normal) (G1 Evacuation Pause) 79M->27M(1024M) 2.929ms`
-Xlog:gc*: `[1.361s][info ][gc          ] GC(0) Pause Young (Normal) (G1 Evacuation Pause) 18874M->9037M(81920M) 23.698ms`

The main difference is that for -Xlog:gc, this is the only line that is printed for an event, whereas -Xlog:gc* prints a CPU time line right afterward, which GCToolkit currently uses as a publish trigger:
`[1.361s][info ][gc,cpu      ] GC(0) User=0.95s Sys=0.09s Real=0.02s`

To address this issue, I've added a Diarizer flag PRINT_CPU_TIMES that is populated when extracting decorators.  If this flag is unset/unsure, the UnifiedG1GCParser.youngDetails() method will populate remaining fields (GC Type - Young, GC Cause extracted from trace 4 (offset -2), StartTime as reported by getClock()) and attempt to publish the event.  I have left in a null check on GCType in the publish condition as a failsafe, but in theory it could be removed.

I think this should be reasonably solid, is a relatively minor change, and it will keep events being published in order.

**Test Updates**
Not much done here, other than adding a couple of updates to existing test cases.  I can provide a sample file, and can update some additional suites, though I wanted to confirm how I can go about doing that.  Is it just a matter of a PR to gctoolkit-testdata with the new file first?

### DIFF
--- a/api/src/main/java/com/microsoft/gctoolkit/jvm/Diary.java
+++ b/api/src/main/java/com/microsoft/gctoolkit/jvm/Diary.java
@@ -378,6 +378,10 @@ public class Diary {
         return isApplicationStoppedTimeKnown() && isApplicationRunningTime();
     }
 
+    public boolean isPrintCPUTimes() {
+    	return isStateKnown(SupportedFlags.PRINT_CPU_TIMES);
+    }
+    
     public void setTimeOfFirstEvent(DateTimeStamp startTime) {
         if ( this.timeOfFirstEvent == null)
             this.timeOfFirstEvent = startTime;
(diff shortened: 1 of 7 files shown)


kcpeppe:
This looks good, thank you! There is additional work that needs to be done but I'll create a new issue to describe that. As for test data, yes, you will need to add the log files to gctoolkit-testdata. Once that is in, I'll trigger a release.

kcpeppe (approved):
Needs a follow on issue to fix diarization tests for the Generational, Z, and Shenandoah collectors.

Issue #433, opened 2025-04-22
Pre-unified G1 Young event not captured - Trace matches are incorrectly offset.


jlittle-ptc:
**Describe the bug**
GC Log generated on Java 1.8.0_442-b06 with -XX:+UseG1GC and -XX:+UseCompressedOops (Latter probably doesn't matter)

The following line does not parse correctly:
`0.867: [GC pause (G1 Evacuation Pause) (young) 52816K->9563K(1024M), 0.0225122 secs]`

PreUnifiedG1GCParser.processYoung is called, which uses an offset of 9 for the match, but the actual start is 12 in this case.  As a result "Unable to calculate generational memory pool summary." is logged when it tries to parse match 9.

**To Reproduce**
Attempt to parse attached file.

**Expected behavior**
GC File is scanned without issue.

[gc.txt](https://github.com/user-attachments/files/19849484/gc.txt)

jlittle-ptc:
Actually, it seems the offsets are just wrong... I tried the other examples provided and all showed the matches at group 12 rather than group 9.  So the additional check around group 9 is unnecessary as it will always trigger the offset. 

Incrementing the offsets in the method by 3 resolves the problem without any test failures.

jlittle-ptc:
PR #438 submitted that fixes the offset issues I was able to identify in pre-unified logging and adds some event format testing to ensure we're parsing correctly.

PR #436, opened 2025-04-24
#433 - Offsets incorrect for Young Events (Pre-unified G1, Generational)

jlittle-ptc:
Two updates were made to fix some offset issues:
* Offsets in GenerationalHeapParser.psYoungNoDetails() incremented by 3 to match current regex.
  * Note: GenerationalHeapParser.psFull() might also have the same problem, but I don't have an example to validate against, so I left it as is and added a comment.
* Offsets in PreUnifiedG1GCParser.processYoung() incremented by 3 to match current regex.

Test cases were added that will fail in the event that the sample lines cannot be processed.  
* There was an existing unit test for PreUnifiedG1GCParser, but it didn't have the assertMemoryPoolValues() helper as it extended com.microsoft.gctoolkit.parser.unittests.ParserTest class instead of com.microsoft.gctoolkit.parser.ParserTest. 
* I created a new class for PreUnifiedG1GCParser tests extending com.microsoft.gctoolkit.parser.ParserTest, using the same general structure as other tests in this directory and added a test.

### DIFF
--- a/parser/src/main/java/com/microsoft/gctoolkit/parser/GenerationalHeapParser.java
+++ b/parser/src/main/java/com/microsoft/gctoolkit/parser/GenerationalHeapParser.java
@@ -1595,15 +1595,17 @@ public class GenerationalHeapParser extends PreUnifiedGCLogParser implements Sim
         publish(collection);
     }
 
+    // #433 - Offset for this call needed to be incremented by 3. (previously 4)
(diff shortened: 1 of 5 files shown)


jlittle-ptc:
I found another offset issue in PreUnifiedG1GC that I can fix with the same PR. (My branch is updated)  

Does it make more sense to cancel this and re-submit after I've gone through my test files and fixed any offset issues I find?

jlittle-ptc:
Need to make a few changes - will resubmit later.

PR #438, opened 2025-04-28
#433 - Fixes and tests for Pre-Unified G1GC and Parallel

jlittle-ptc:
**Fixes:**
* GenerationalHeapParser.psYoungNoDetails() - Offsets were off by 3.
  * Note: I didn't have any sample lines for a full GC, but I suspect the same issue will apply there.  I did not make the change at this time, but left a note in a comment.
* PreUnifiedG1GCParser.processYoungGenCollection() - Offset for initial-mark was off by 3.

**Test Cases:**
* Event parsing test cases for Pre-unified G1 and Parallel/CMS with Details/NoDetails

### DIFF
--- a/IT/pom.xml
+++ b/IT/pom.xml
@@ -5,7 +5,7 @@
     <parent>
         <groupId>com.microsoft.gctoolkit</groupId>
         <artifactId>gctoolkit</artifactId>
-        <version>3.6.0-SNAPSHOT</version>
+        <version>3.7.0-SNAPSHOT</version>
         <relativePath>../pom.xml</relativePath>
     </parent>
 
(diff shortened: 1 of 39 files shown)


kcpeppe:
Triggering a System.gc() via jcmd (or JMX or ...) would yield the missing Full GC record.

dsgrieve:
@jlittle-ptc - Do you have a gclog that exhibits this behavior? If so, can you attach it here and we can add it to the gclogs. Thanks.

jlittle-ptc:
> @jlittle-ptc - Do you have a gclog that exhibits this behavior? If so, can you attach it here and we can add it to the gclogs. Thanks.

[gc.txt](https://github.com/user-attachments/files/19957775/gc.txt)

This file was attached to #433, where the issue was reported. 

Once I get closer to finishing my review, I'll look to upload any examples that I have that are different than what's in the test data.

kcpeppe (approved):

Issue #336, opened 2024-03-11
AbstractJavaVirtualMachine.getCommandLine() is currently stubbed out, can't retrieve CLI parameters for Pre-Unified log.


Bluetopia:
**Describe the bug**
The getCommandLine() method has been stubbed out with a TODO.  See https://github.com/microsoft/gctoolkit/blob/1476fc128465991f2c531645c780d3ed7c865f2c/api/src/main/java/com/microsoft/gctoolkit/jvm/AbstractJavaVirtualMachine.java#L84

**Expected behavior**
Expected this method to return the command line string from a Pre-Unified log file. 

**Additional context**
There is a pending issue for building the Command Line args for Unified logging here: https://github.com/microsoft/gctoolkit/issues/263 which states that retrieving the command line is supported for Pre-Unified log files.  However, the AbstractJavaVirtualMachine.getCommandLine() method is not overridden in PreUnifiedJavaVirtualMachine.  Not sure if this was stubbed out while waiting for issue 263 or an oversight.

karianna:
@kcpeppe & @dsgrieve for comment

kcpeppe:
This is behaviour that we had in Censum but didn't get transferred to GCToolKit. It is a lower priority item for me ATM.

PR #440, opened 2025-04-30
Fix for Legacy ZGC log files without details (-Xlog:gc)

jlittle-ptc:
[zgc-noGen-java21-noDetail.log](https://github.com/user-attachments/files/19978052/zgc-noGen-java21-noDetail.log)

Small log sample attached from Java 21 without +ZGenerational and using -Xlog:gc rather than -Xlog:gc* (detailed)

ZGCParser was unable to parse the events from this log as there is only one line logged per event, and thus, no "cycle start" that would generate the forwardReference.  This is not an issue for Generational, as two lines are logged for each event, so the special case only applies to Full/Garbage type ZGC events.

Fixed this issue and added test cases for both Generational and non-generational lines without details to ZGCParserTest.

Also took the opportunity to update deprecated .getTimeStamp() calls to .toSeconds() in ZGCParserTest.

### DIFF
--- a/IT/pom.xml
+++ b/IT/pom.xml
@@ -5,7 +5,7 @@
     <parent>
         <groupId>com.microsoft.gctoolkit</groupId>
         <artifactId>gctoolkit</artifactId>
-        <version>3.6.0-SNAPSHOT</version>
+        <version>3.7.0-SNAPSHOT</version>
         <relativePath>../pom.xml</relativePath>
     </parent>
 
--- a/api/pom.xml
+++ b/api/pom.xml
@@ -5,7 +5,7 @@
     <parent>
         <groupId>com.microsoft.gctoolkit</groupId>
         <artifactId>gctoolkit</artifactId>
-        <version>3.6.0-SNAPSHOT</version>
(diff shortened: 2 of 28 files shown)


dsgrieve:
@kcpeppe - I'm thinking we want to handle this is a consistent way across all collectors. We had the notion of using the diarizer to know whether the log didn't have enough detail. We discussed this in the (somewhat stale) PR #409

jlittle-ptc:
I was debating taking a stab at #336 and implementing something like that where the diary would contain information about the Java Flags used from the command line if present (pre-unified) or what we can infer from unified logs during the diarizing process.  I was thinking of adding a key/value map for storing this information rather than exclusively using flags as is currently done.  

Unfortunately, my priorities are going to be shifting for a bit, so I'm not certain if I can get to it in the near term.  But if you have some thoughts, I'm all ears.

karianna:
@jlittle-ptc apologies for not pushing on this earlier.  We did merge another major refactor of the ZGC parser etc, so I'm afraid this will need a rebase.

Issue #442, opened 2025-05-13
Exception when using RotatingGCLogFile


srilakshmikanthanp:
I would like to extend my sincere thanks for developing and maintaining the GC Toolkit. I am encountering a DateTimeParseException when trying to use the RotatingGCLogFile class from the GC Toolkit to analyze GC log files. 

~~~java
package com.example;

import java.io.IOException;
import java.nio.file.Path;

import com.microsoft.gctoolkit.GCToolKit;
import com.microsoft.gctoolkit.io.RotatingGCLogFile;

public class Main {
    public static void main(String[] args) throws IOException {
        var logFile = new RotatingGCLogFile(Path.of("./logs"));
        var gcToolKit = new GCToolKit();
        var jvm = gcToolKit.analyze(logFile);
    }
}
~~~

Above throws exception,

~~~shell
Main 
Exception in thread "main" java.time.format.DateTimeParseException: Text '[2025-05-13T14:47:20.320+0530]' could not be parsed at index 0
        at java.base/java.time.format.DateTimeFormatter.parseResolved0(DateTimeFormatter.java:2052)
        at java.base/java.time.format.DateTimeFormatter.parse(DateTimeFormatter.java:1880)
        at com.microsoft.gctoolkit.time.DateTimeStamp.dateFromString(DateTimeStamp.java:52)
        at com.microsoft.gctoolkit.time.DateTimeStamp.fromGCLogLine(DateTimeStamp.java:92)
        at java.base/java.util.stream.ReferencePipeline$3$1.accept(ReferencePipeline.java:197)
        at java.base/java.util.ArrayList$ArrayListSpliterator.forEachRemaining(ArrayList.java:1625)
        at java.base/java.util.stream.AbstractPipeline.copyInto(AbstractPipeline.java:509)
        at java.base/java.util.stream.AbstractPipeline.wrapAndCopyInto(AbstractPipeline.java:499)
        at java.base/java.util.stream.ReduceOps$ReduceOp.evaluateSequential(ReduceOps.java:921)
        at java.base/java.util.stream.AbstractPipeline.evaluate(AbstractPipeline.java:234)
        at java.base/java.util.stream.ReferencePipeline.reduce(ReferencePipeline.java:662)
        at java.base/java.util.stream.ReferencePipeline.max(ReferencePipeline.java:698)
        at com.microsoft.gctoolkit.io.GCLogFileSegment.ageOfJVMAtLogEnd(GCLogFileSegment.java:146)
[...]

PR #444, opened 2025-05-29
Ignore [] brackets while capturing date with UNIFIED_DATE_TIMESTAMP

srilakshmikanthanp:
Hi,
I made a small fix that addresses the `DateTimeParseException` (#442) when parsing GC logs with RotatingGCLogFile. I’m not deeply familiar with GC internals or GC log formats, so this change is based on my understanding of the related code.
Please review and advise if any improvements are needed.
Thanks!


kcpeppe:
Can you provide a matching test case into DateTimeStampTest?

srilakshmikanthanp:

Issue #445, opened 2025-06-04
Generational ZGC incorrect handling of multiple concurrent collections


akiselev98:
Hi. #414 added generational ZGC support, however, there are some corner cases that are currently handled incorrectly.

1. It assumes that a major cycle contains exactly 1 young collection, and exactly 1 old collection. This is not the case. A major can have exactly 1 old collection, and *at least* one young collection (I have seen examples with 2 young collections in a single major cycle, I don't know if it's possible to have 3 or more)
2. The parser has a single forward reference. If a major cycle starts, and then a minor cycle starts before that major cycle has finished, then the parser will discard the information from the major cycle

Example:
```
❯ egrep -e '\[gc,phases.+(Young|Old) Generation' gc.log
[2024-11-19T09:48:31.967-0600][info][gc,phases   ] GC(0) Y: Young Generation
[2024-11-19T09:48:31.984-0600][info][gc,phases   ] GC(0) Y: Young Generation 88M(1%)->54M(0%) 0.017s
[2024-11-19T09:48:31.984-0600][info][gc,phases   ] GC(0) O: Old Generation
[2024-11-19T09:48:31.997-0600][info][gc,phases   ] GC(0) O: Old Generation 54M(0%)->56M(0%) 0.013s
[2024-11-19T09:48:38.557-0600][info][gc,phases   ] GC(1) Y: Young Generation
[2024-11-19T09:48:39.641-0600][info][gc,phases   ] GC(1) Y: Young Generation 828M(5%)->1392M(8%) 1.084s
[2024-11-19T09:48:39.641-0600][info][gc,phases   ] GC(1) O: Old Generation
[2024-11-19T09:48:39.698-0600][info][gc,phases   ] GC(1) O: Old Generation 1392M(8%)->1506M(9%) 0.056s
[2024-11-19T09:48:39.753-0600][info][gc,phases   ] GC(2) Y: Young Generation
[2024-11-19T09:48:42.252-0600][info ][gc,phases   ] GC(2) Y: Young Generation 1642M(10%)->7986M(49%) 2.499s
[2024-11-19T09:48:42.252-0600][info ][gc,phases   ] GC(2) O: Old Generation
[2024-11-19T09:48:42.498-0600][info ][gc,phases   ] GC(2) O: Old Generation 8002M(49%)->8920M(54%) 0.246s
[2024-11-19T09:48:42.508-0600][info ][gc,phases   ] GC(3) Y: Young Generation
[...]

PR #447, opened 2025-06-11
Report ZGC individual collections instead of cycles (#445)

akiselev98:
Previously, the ZGC parser would attempt to construct a single event for each minor/major cycle, with each minor cycle containing a single young collection, and each major cycle containing a single young and a single old collection.

This created a few problems.

First, the implementation of the parser assumed that only one cycle could be active at any given time. As a result, any time a minor cycle started before the previous major cycle was finished, all data collected about the major cycle would be lost.

Second, the data model for a MajorZGCCycle assumed there could be only one young collection per cycle. This is not the case. There can be (and frequently are) multiple young collections in a major cycle.

Even if the implementation and datamodel were fixed to account for the concurrent nature of generational ZGC, grouping collections by cycle would still create API usability problems. Individual collections within a cycle independently report various JVM/heap level metrics at specific points in time. Within a single major cycle, those points in time may be very far apart, and overlap with a large number of minor cycles. If an end user wished to construct a timeline for a particular metric (for example, heap occupancy), they would need to traverse all the reported cycles, extract the metrics from the individual collections, and put them in order.

To address the above issues, the parser and data models are refactored to report each individual collection (young, old, or full) immediately once they are completed instead of waiting until the parent cycle is done. The parser maintains 3 independent forward references (one for each collection type: old, young, and full). As a result, the parser will not lose state even when lines from an old collection are intermixed with lines from a young collection.


karianna:
@dsgrieve / @kcpeppe - in case you wanted to comment - else I'll merge this in 42-48 hours.

dsgrieve (commented):
LGTM. This change does break API.

karianna:
> LGTM. This change does break API.

Fair, would require version bump if we're being semantic

parser/src/test/java/com/microsoft/gctoolkit/parser/ZGCParserTest.java
```diff
@@ -3,14 +3,7 @@
...
 import com.microsoft.gctoolkit.event.jvm.JVMEvent;
-import com.microsoft.gctoolkit.event.zgc.FullZGCCycle;
-import com.microsoft.gctoolkit.event.zgc.ZGCAllocatedSummary;
-import com.microsoft.gctoolkit.event.zgc.ZGCGarbageSummary;
-import com.microsoft.gctoolkit.event.zgc.ZGCLiveSummary;
-import com.microsoft.gctoolkit.event.zgc.ZGCMemoryPoolSummary;
-import com.microsoft.gctoolkit.event.zgc.ZGCMemorySummary;
-import com.microsoft.gctoolkit.event.zgc.ZGCMetaspaceSummary;
-import com.microsoft.gctoolkit.event.zgc.ZGCReclaimSummary;
+import com.microsoft.gctoolkit.event.zgc.*;
```
karianna, line 6: Prefer explicit imports

akiselev98: Ah, I thought I had caught all the places where IntelliJ did that, but apparently not. Will fix.

PR #458, opened 2025-10-17
Optimize decorator parsing

akiselev98:
Profiling showed that a relatively high percentage of CPU time and memory allocations are coming from decorator parsing. This makes sense, since this parsing is applied to every unified log line which goes through gctoolkit.

To speed things up, we:
- Fold the tags regex into the broader decorator regex, avoiding the need for a second matcher, and removing the need for an expensive negative lookbehind in the tag pattern
- Add a start of line anchor to the decorator regex
- Defer sanitization of tags until getTags() is called for the first time

To reduce unnecessary memory allocations, match groups are retrieved once and stored in an array.

In testing, these optimizations cut the running time of gctoolkit roughly in half.

### DIFF
--- a/parser/src/main/java/com/microsoft/gctoolkit/parser/jvm/Decorators.java
+++ b/parser/src/main/java/com/microsoft/gctoolkit/parser/jvm/Decorators.java
@@ -64,16 +64,16 @@ public class Decorators {
     private static final long TWENTY_YEARS_IN_MILLIS = 731L * 24L * 60L * 60L * 1000L;
     private static final long TWENTY_YEARS_IN_NANO = 731L * 24L * 60L * 60L * 1000L;
 
-    private static final int DATE_STAMP_GROUP = 1;
-    private static final int UPTIME_GROUP = 2;
-    private static final int TIME_MILLIS_OR_MAYBE_UPTIME_MILLIS_GROUP = 3;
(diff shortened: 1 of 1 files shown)


parser/src/main/java/com/microsoft/gctoolkit/parser/jvm/Decorators.java
```diff
@@ -66,52 +66,51 @@ public class Decorators {
...
 
-        for ( int i = 1; i <= decoratorMatcher.groupCount(); i++) {
-            if ( decoratorMatcher.group(i) != null)
+        // Retrieving a group from a matcher calls substring each time
+        // Store all the groups in an array ahead of time to avoid paying this cost unnecessarily
+        decoratorGroups = new String[11];
+        for (int i = 1; i <= decoratorMatcher.groupCount(); i++) {
+            String group = decoratorMatcher.group(i);
+            decoratorGroups[i] = group;
+            if ( group != null)
```
karianna:
```suggestion
            if (group != null)
```

karianna (dismissed):

obourgain:
this is indeed faster, about twice faster for my use case too.

johnoliver (approved):

karianna (approved):

PR #459, opened 2025-11-03
Heap region size for G1GC is no longer parsed after JDK-8244817

fthevenet:
Starting with [JDK-8244817](https://bugs.openjdk.org/browse/JDK-8244817), the heap region size can no longer be detected by the pattern `Heap region size` but by `Heap Region Size` (note the capitalized `Region` and `Size`) in G1GC log files.

As a result of not being able to retrieve heap region size, GCToolkit is not able to report the occupancy of the various  G1 memory pool types (eden, survival, old).


karianna (approved):

karianna:
> Thanks @karianna Do you have an estimate you can share for when the next release is scheduled?
> 
> Cheers!

Over to @d3r3kk & @johnoliver to schedule that in with the team

PR #465, opened 2025-11-13
Faster Decorators parsing

obourgain:
This is pretty hot in profiles.

I know it was already improved in https://github.com/microsoft/gctoolkit/pull/458, but we can get rid of the regex altogether.

### DIFF
--- a/parser/src/main/java/com/microsoft/gctoolkit/parser/jvm/Decorators.java
+++ b/parser/src/main/java/com/microsoft/gctoolkit/parser/jvm/Decorators.java
@@ -3,17 +3,16 @@
 package com.microsoft.gctoolkit.parser.jvm;
 
 import com.microsoft.gctoolkit.parser.unified.UnifiedLoggingLevel;
-import com.microsoft.gctoolkit.parser.unified.UnifiedLoggingTokens;
 import com.microsoft.gctoolkit.time.DateTimeStamp;
 
+import java.time.ZoneOffset;
 import java.time.ZonedDateTime;
-import java.time.format.DateTimeFormatter;
-import java.time.temporal.TemporalAccessor;
 import java.util.Arrays;
+import java.util.List;
 import java.util.Optional;
 import java.util.logging.Level;
 import java.util.logging.Logger;
-import java.util.regex.Matcher;
+import java.util.regex.Pattern;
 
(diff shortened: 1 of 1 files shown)


parser/src/main/java/com/microsoft/gctoolkit/parser/jvm/Decorators.java
```diff
@@ -60,6 +59,11 @@ public class Decorators {
 
     private static final Logger LOGGER = Logger.getLogger(Decorators.class.getName());
 
+    // cheap optim, assume that most logs are at info or debug level and thus put those at the head of the list
```
karianna:
```suggestion
    // Predictive optimization, assume that most logs are at info or debug level and thus put those at the head of the list
```

parser/src/main/java/com/microsoft/gctoolkit/parser/jvm/Decorators.java
```diff
@@ -60,6 +59,11 @@ public class Decorators {
...
+    private static final List<String> LEVELS = List.of("info", "debug", "error", "warning", "trace", "develop");
+    // not the same as UnifiedLoggingTokens.TAGS as there is no literal brackets here
```
karianna:
```suggestion
    // This is *not* the same as UnifiedLoggingTokens.TAGS as there are no literal brackets here
```

parser/src/main/java/com/microsoft/gctoolkit/parser/jvm/Decorators.java
```diff
@@ -75,44 +79,175 @@ public class Decorators {
...
 
     private void extractValues(String line) {
-        if (!line.startsWith("["))
+        if (line.charAt(0) != '[') {
             return;
-
-        Matcher decoratorMatcher = UnifiedLoggingTokens.DECORATORS.matcher(line);
-        if (!decoratorMatcher.find()) {
+        }
+        // don't assign fields until we are done parsing, to avoid inconsistent state
```
karianna:
```suggestion
        // To avoid inconsistent state, don't assign fields until we are done parsing.
```

karianna (dismissed):
@johnoliver Over to you

karianna (approved):

PR #470, opened 2025-11-23
Fixed unsightly capitalization in some GarbageCollectionTypes labels

fthevenet:
This is a trivial PR that fixes incorrect capitalization on two labels of the `GarbageCollectionTypes` enum.


karianna (approved):

karianna:
I'm happy to keep this merged as is.  I think if we find a log in the wild with the 'bad' capitalization we can add it to the test suite and cover it.

PR #579, opened 2026-08-12
Cache duration Pattern in GenerationalHeapParser CMS remark path

### LINKED ISSUE #567: Parser bug: Pattern.compile() called per-invocation in GenerationalHeapParser.splitRemarkReferenceWithWeakReferenceSplitBug
## Describe the bug

`GenerationalHeapParser.splitRemarkReferenceWithWeakReferenceSplitBug(GCLogTrace, String)` compiles a regex **on every call** instead of caching it in a `static final` field. Regex compilation on a hot parse path is pure overhead and shows up clearly in allocation profiling.

`parser/src/main/java/com/microsoft/gctoolkit/parser/GenerationalHeapParser.java` (line ~1920):
```java
public void splitRemarkReferenceWithWeakReferenceSplitBug(GCLogTrace trace, String line) {
    GCLogTrace remarkTrace = REMARK_CLAUSE.parse(line);
    Pattern durationGroupPattern = Pattern.compile(".* " + PAUSE_TIME);  // recompiled every call
    Matcher matcher = durationGroupPattern.matcher(line);
    ...
```

Every other `Pattern.compile` in the parser (e.g. `AbstractLogTrace`, `ParNewSerial`, `PreUnifiedDiarizer`) is correctly a `static final` constant — this call site is the outlier.

## Evidence (JFR `settings=profile`, ~28s run)

Workload: sample `Main` parsing a 195 MB JDK8 ParallelGC rolling log (`gclogs/rolling/jdk8/aagl_prd/gc.log`), `-Xmx1500m`, G1, Java 25 (GraalVM 25.0.3). ~19,985 MB sampled allocations.

Regex **compilation** (not matching) is visibly present in the allocation-by-site profile:

| Alloc site | Share |
|---|---|
| `java.util.regex.Pattern.compile()` | 6.83% |
| `java.util.regex.Pattern.RemoveQEQuoting()` (called from compile) | 6.87% |

Compilation should be a one-time startup cost, not ~7% of steady-state allocation.

## Expected behavior

Patterns compiled once and reused:
```java
private static final Pattern DURATION_GROUP_PATTERN = Pattern.compile(".* " + PAUSE_TIME);
```

## Additional context

- Audit sibling methods for the same anti-pattern (any in-method `Pattern.compile`). The confirmed offending line is `GenerationalHeapParser:1920`.
- The sample log used here is ParallelGC, so this specific CMS-remark method is not the dominant contributor in this run — but the recompile is still incorrect and will hurt on CMS logs.
- Environment: macOS (Darwin 25.5.0, arm64), Java HotSpot 25.0.3+9-LTS (GraalVM), GCToolkit 3.7.1-SNAPSHOT.

Hashim1999164:
## Summary

`splitRemarkReferenceWithWeakReferenceSplitBug` recompiled `Pattern.compile(".* " + PAUSE_TIME)` on every call. That matches other parser patterns that already use `static final` compiled patterns.

This change introduces `DURATION_GROUP_PATTERN` and reuses it for matching.

Fixes #567

## Test plan

1. `./mvnw -pl parser -am -DskipTests compile`
2. Optional: run parser tests that cover CMS remark / weak reference split bug lines

### DIFF
--- a/parser/src/main/java/com/microsoft/gctoolkit/parser/GenerationalHeapParser.java
+++ b/parser/src/main/java/com/microsoft/gctoolkit/parser/GenerationalHeapParser.java
@@ -65,7 +65,8 @@ public class GenerationalHeapParser extends PreUnifiedGCLogParser implements Sim
     private static final Logger LOGGER = Logger.getLogger(GenerationalHeapParser.class.getName());
 
     // Cached for the CMS remark/weak-reference split-bug path; avoid recompiling on every parse call.
-    private static final Pattern DURATION_GROUP_PATTERN = Pattern.compile(".* " + PAUSE_TIME);
+    // [^\n]* preserves last-match semantics of greedy .* without a ReDoS hotspot.
+    private static final Pattern DURATION_GROUP_PATTERN = Pattern.compile("[^\\n]* " + PAUSE_TIME);
 
     private ParNew parNewForwardReference;
     private GarbageCollectionTypes garbageCollectionTypeForwardReference;
@@ -629,9 +630,11 @@ public class GenerationalHeapParser extends PreUnifiedGCLogParser implements Sim
         gcCauseForwardReference = GCCause.PROMOTION_FAILED;
         ArrayList<Integer> blocks = new ArrayList<>();
         GCLogTrace block = PARNEW_PROMOTION_FAILURE_SIZE_BLOCK.parse(line);
-        do {
-            blocks.add(block.getIntegerGroup(1));
-        } while (block.hasNext());
+        if (block != null) {
+            do {
+                blocks.add(block.getIntegerGroup(1));
+            } while (block.hasNext());
+        }
         promotionFailureSizesForwardReference = new int[blocks.size()];
         for (int index = 0; index < blocks.size(); index++)
             promotionFailureSizesForwardReference[index] = blocks.get(index);
@@ -1820,7 +1823,8 @@ public class GenerationalHeapParser extends PreUnifiedGCLogParser implements Sim
     private void precleanTimedoutWithCards(GCLogTrace trace, String line) {
         abortPrecleanDueToTime = true;
         GCLogTrace concurrentPhase = new GCParseRule("X",CMS_PHASE_END).parse(line);
-        endOfConcurrentPhase(concurrentPhase, concurrentPhase.getDateTimeStamp(), 0);
+        if (concurrentPhase != null)
+            endOfConcurrentPhase(concurrentPhase, concurrentPhase.getDateTimeStamp(), 0);
     }
 
     private void shouldCollectConcurrent(GCLogTrace trace, String line) {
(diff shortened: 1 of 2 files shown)


parser/src/main/java/com/microsoft/gctoolkit/parser/GenerationalHeapParser.java
```diff
@@ -64,6 +64,10 @@ public class GenerationalHeapParser extends PreUnifiedGCLogParser implements Sim
 
     private static final Logger LOGGER = Logger.getLogger(GenerationalHeapParser.class.getName());
 
+    // Cached for the CMS remark/weak-reference split-bug path; avoid recompiling on every parse call.
+    // [^\n]* preserves last-match semantics of greedy .* without a ReDoS hotspot.
+    private static final Pattern DURATION_GROUP_PATTERN = Pattern.compile("[^\\n]* " + PAUSE_TIME);
```
karianna: Could we move this pattern into `CMSPatterns` instead of keeping it in `GenerationalHeapParser`?

This expression is specific to the CMS split-remark/weak-reference parsing path, and `CMSPatterns` already contains the related `SPLIT_REMARK_REFERENCE_BUG` rule. Keeping both patterns together would make the ownership clearer and avoid mixing parser state with CMS grammar definitions.

I’d suggest defining it there as a named `GCParseRule`, for example:

```java
GCParseRule SPLIT_REMARK_REFERENCE_BUG_DURATION =
        new GCParseRule("SPLIT_REMARK_REFERENCE_BUG_DURATION", "[^\\n]* " + PAUSE_TIME);
```

Then this method can call `parse(line)` and use `GCLogTrace#getPauseTime()`, consistently with the rest of the parser.

karianna (changes requested):

parser/src/main/java/com/microsoft/gctoolkit/parser/GenerationalHeapParser.java
Hashim1999164: Moved the duration pattern into CMSPatterns as SPLIT_REMARK_REFERENCE_BUG_DURATION. The parser now uses parse() and getPauseTime().

Hashim1999164: @karianna the duration pattern now lives in CMSPatterns as SPLIT_REMARK_REFERENCE_BUG_DURATION and the parser uses parse() plus getPauseTime(). Please take another look when you have a chance.

skrcode:
I ran JAIPilot Cloud against this exact PR head. It found one additional internal allocation cleanup in the changed ParNew promotion-failure path: collect sizes in a growable primitive int array instead of boxing each value into an ArrayList<Integer> and then unboxing into the final array.

Tests pin zero, one, and 28 captured blocks and their exact order. The same focused and 124 parser tests passed before and after; the seven-module reactor passed. A fixed 28-block collection fixture eliminated 28 boxed Integers plus ArrayList growth, with a five-trial isolated median of 733.82 ms to 537.30 ms over 5,000,000 iterations.

PR directly onto this source branch: https://github.com/Hashim1999164/gctoolkit/pull/1  
Cloud-generated draft and full evidence: https://github.com/skrcode/gctoolkit/pull/2

Regex parsing dominates real log processing and the populated field currently has no downstream reader, so this is an optional internal allocation cleanup rather than a broad throughput claim. Microsoft CLA handling may require folding the change into the contributor commit.

karianna (approved):

karianna:
@johnoliver Can you approve and merge

PR #588, opened 2026-09-03
Record and report safepoint information from logs

AlinaHovakimyan:
On ZGC logs example i saw that Safepoint lines were not getting reported by gctoolkit. The pattern existed, but was never used. So I have added parsing and reporting of it via ApplicationStoppedTimes.

- I have added fromName() function to VMOperation's enum, as in every release vm operations get added or deleted, and by this gctoolkit will return null rather than throwing for unknowns, so that an unrecognized operation costs the reason and not the whole event.
- Safepoint line differs between jdk versions, so i added optional parts to the pattern: cleanup -jdk17, jdk21, Leaving safepoint - jdk21, Threads - jdk25

- Also added tests and tested locally manually


AlinaHovakimyan:
@karianna, @d3r3kk  can you please review this MR, i am not able to add reviewers

kcpeppe:
For historical reasons, Safepoint is captured in it's own parser (SafepointParser), not in the individual parsers for each collector. There are examples on how to capture safepoints. That said, the formats for parsing haven't been kept up to date.

AlinaHovakimyan:
> For historical reasons, Safepoint is captured in it's own parser (SafepointParser), not in the individual parsers for each collector. There are examples on how to capture safepoints. That said, the formats for parsing haven't been kept up to date.

@kcpeppe thank you very much for checking the MR.  One thing to confirm, since SafepointParser extends PreUnifiedGCLogParser, I'm planning to add UnifiedSafepointParser extending UnifiedGCLogParser, registered in  the same pre-unified/unified split JVMEventParser, SurvivorMemoryPoolParser and other parsers have, and not widen the       SafepointParser itself. It'll publish Safepoint on EventSource.SAFEPOINT, with the JDK 14+ phases added to that class alongside the existing  PrintSafepointStatistics fields, following how G1GCPauseEvent carries unified-only region summaries with a sentinel for the  pre-unified case.
Are you fine with this approach?

kcpeppe:
Odd, I don't see a sentinel in G1GCPauseEvent. At any rate, mixing pre-unified with unified wouldn't be something I'd recommend doing. Theory being, it should be possible to remove pre-unified without leaving any trace of it behind and without breaking unified. So, pre-unified and unified share but only what is common to both. As for safepoints, they are treated differently than GC events for historical reasons that have long since gone away. I had ideas for how to integrate safepoint events into the overall event hierarchy in a more sensible way but life events parked this effort. Anyways, not my call but to me what you're proposing is consistant with how things are currently structured.

AlinaHovakimyan:
@kcpeppe thank you for the confirmation, I will move forward with this.

> Odd, I don't see a sentinel in G1GCPauseEvent. 

I was reffering to RegionSummary in G1GCPauseEvent, which is getting constructed only in UniefiedG1GCParser, but even if this is the case, one mismatch with the approach does not mean we have to allow new ones.

kcpeppe:
@AlinaHovakimyan , do you mind pointing to this sentinal by providing a class::method::line#? I'm not finding it. AFAIK, there is no mixing a pre-unified and unified events. There is sharing in the hierarchy but that sharing is limited to things that are sharable. There should be no denial of inheritance or other like antipatterns in the code base.

AlinaHovakimyan:
> [AlinaHovakimyan](https://github.com/AlinaHovakimyan) , do you mind pointing to this sentinal by providing a class::method::line#? I'm not finding it. AFAIK, there is no mixing a pre-unified and unified events. There is sharing in the hierarchy but that sharing is limited to things that are sharable. There should be no denial of inheritance or other like antipatterns in the code base.

@kcpeppe sure, the [getEdenRegionSummary](https://github.com/microsoft/gctoolkit/blob/765a084b5f946bd5830395d0c19d0c4072ab9bdf/api/src/main/java/com/microsoft/gctoolkit/event/g1gc/G1GCPauseEvent.java#L84)() in G1GCPauseEvent.java returns 
the NULL_REGION([line](https://github.com/microsoft/gctoolkit/blob/765a084b5f946bd5830395d0c19d0c4072ab9bdf/api/src/main/java/com/microsoft/gctoolkit/event/g1gc/G1GCPauseEvent.java#L17)) or the edenRegion if its not null. The only call that populates this field is G1GCPauseEvent::addRegionSummary ([line](https://github.com/microsoft/gctoolkit/blob/765a084b5f946bd5830395d0c19d0c4072ab9bdf/api/src/main/java/com/microsoft/gctoolkit/event/g1gc/G1GCPauseEvent.java#L76)), which is called from G1GCForwardReference.java:629, which is used solely by UnifiedG1GCParser. PreUnifiedG1GCParser never touches them.

kcpeppe:
@AlinaHovakimyan , ok, that would be a design bug. One of the design decisions made early on was to separate the unstable log formats from the stable analysis by converting the information in the logs to the events. The decision did ease the pain of maintaining parse rules while keeping everything else stable. It even mostly worked when Unified Logging was introduced in 9. It broke a lot of things but not everything. My guess is laziness ruled over creating a new event for this G1 phase. It's likely too disruptive to "fix" that now so I'd doc it and leave it. That it differentiates UL from pre-UL is accidential. I'd rely on the information in the diary to make that determination.

AlinaHovakimyan:
@kcpeppe I have modified the approach to mirror SafepointParser and introduce UnifiedSafepointParser, can you please check this out again?
Two notes:
- SafepointParser was only listed in module-info's provides clause, never in the hardcoded fallback list in GCToolKit that's used when running off the classpath. If i read it correctly, that meant pre-unified safepoint parsing silently did nothing outside the module path. I have added both parsers to the list. 
- UnifiedGCLogParser had to become public so the new parser can extend it from the parser.vmops  subpackage. Happy to move the parser into the parser package instead if you'd rather not widen that. But I just added to the folder where SafepointParser is.

AlinaHovakimyan:
@kcpeppe @d3r3kk @karianna @johnoliver can you please check this MR? let me know if the review process changed and i should trigger some AI review before pinging you. Thanks in advance

